PR #4, opened 2019-01-21
Issue#1

bryceosterhaus:
@jbalsas Let me know if I may have missed something. Only area of concern I have is the hard coding of [this line](https://github.com/bryceosterhaus/liferay-npm-tools/compare/issue%231?expand=1#diff-84038447b28be55412923794117649b3R10). I wasn't sure of another way around this. Let me know if you have any ideas.

Here is also a sample of the output failure [gist](https://gist.github.com/bryceosterhaus/c15cf831821dd9b7240b911bcf9eda45)

### DIFF
--- a/packages/liferay-jest-junit-reporter/index.js
+++ b/packages/liferay-jest-junit-reporter/index.js
@@ -2,6 +2,18 @@
 
 const xml = require('xml');
 const fs = require('fs');
+const path = require('path');
+const stripAnsi = require('strip-ansi');
+
+const NEW_LINE = '\n';
+
+const RELATIVE_APPS_DIR = '/modules/apps/';
+
+const APPS_DIR = new RegExp('^.+' + RELATIVE_APPS_DIR);
+
+function formatDirectoryPath(dirPath) {
+	return dirPath.replace(APPS_DIR, '').replace(/\//g, '.');
+}
 
 module.exports = report => {
 	const generalMetrics = {
@@ -11,7 +23,7 @@ module.exports = report => {
 			errors: 0,
 			failures: report.numFailedTests,
 			name: 'Jest',
-			package: '',
+			package: formatDirectoryPath(process.cwd()),
 			skipped: 0,
 			tests: report.numTotalTests,
 			time: 0,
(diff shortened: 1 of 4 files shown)


packages/liferay-jest-junit-reporter/index.js
```diff
@@ -2,41 +2,95 @@
...
-
-  fs.writeFileSync('TEST-frontend-js.xml', xml(
-    {'testsuite': [generalMetrics, ...testResults]},
-    { declaration: true, indent: '  '}
-  ));
-
-  return report;
+const NEW_LINE = '\n';
+
+const RELATIVE_APPS_DIR = '/modules/apps/';
```
jbalsas: I think this should be fine for now, since this is mostly to be consumed by Portal modules. To ease your concerns a little bit... maybe we could make this a configuration option, being `/modules/apps/` the default? That way we protect for future usage if necessary?

packages/liferay-jest-junit-reporter/index.js
```diff
@@ -2,41 +2,95 @@
...
+}
+
+module.exports = report => {
+	const generalMetrics = {
+		_attr: {
+			hostname: '',
+			id: 0,
+			errors: 0,
+			failures: report.numFailedTests,
+			name: 'Jest',
```
jbalsas: I know this was here way before this, so I'm just pointing it out for the future. Having a hardcoded `Jest` name here will group all tests under a `Jest` hierarchy in the Jenkins reports. See attached image:

screen shot 2019-01-22 at 10 13 06

@john-co, what do you think about this? Wouldn't it be better if we could avoid this additional level and just show the module there, for example?

packages/liferay-jest-junit-reporter/index.js
```diff
@@ -2,41 +2,95 @@
...
+						classname: formatDirectoryPath(
+							path.dirname(testCase.testFilePath)
+						),
+						name: testCase.fullName,
+						time: testCase.duration / 1000
+					}
+				}
+			];
+
+			if (testCase.failureMessages && testCase.failureMessages.length) {
```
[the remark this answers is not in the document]
bryceosterhaus: Good catch, I didnt think about the contents of the errors. Ill add that in.

packages/liferay-jest-junit-reporter/index.js
john-co: @jbalsas, hardcoding the name there does not control the package name. This will be fixed with the added commits.

**For detailed explanation on modifying package name:**
The package name itself is controlled by trimming off what comes after the last `.` in the classname value in each testcase. In this case Management Toolbar TEST-frontend-js.xml had `Jest.ManagementToolbar` as classname value so jenkins takes Jest as the package name and uses ManagementToolbar as the child in the navigation structure. If the classname value does not have a `.` then it defaults to `root` as package name.

**To see an example of what this updated jest-reporter will produce:**
Using the generated TEST*.xml file https://gist.github.com/bryceosterhaus/c15cf831821dd9b7240b911bcf9eda45, it will produce the following results on PR tester:
https://github.com/john-co/liferay-portal/pull/49#issuecomment-456499485
And on jenkins:
https://test-1-22.liferay.com/job/test-portal-acceptance-pullrequest-batch(master)/135603/AXIS_VARIABLE=0,label_exp=!master/testReport/

selection_058

Let me know if this sample output will be helpful or if you'd like to see other modifications.

**Note:** I included a manually modified TEST-frontend-js.xml file for managmeent toolbar. This is ideally what I'd like the PR tester output to show.

jbalsas: Sounds good, John, thanks for the explanation!

Based on the screenshot... couldn’t we do something so the testnames are a bit more readable?

jbalsas: Maybe we could shorten the packagename if that’s expected to be there... I don’t remember such long testNames or packageNames in the Java unit ones, though granted, I don’t pay that much attention to them 😅

john-co: In the java tests that I looked at, they actually are that long! In that screen shot the Test Name includes the packageName. But if you navigate through the links for package and then click the classname, you'll find the test names trimmed down.

Here's the view if you click the package `dynamic-data-mapping.dynamic-data-mapping-form-field-type.test.metal.js`
selection_059

And then if you click class `KeyValue`
selection_060

john-co: It's only for the `For All Failed Tests` summary section is where it lists out the test's full path.

john-co: If any further consolation is needed, the PR tester links will only show the class and test name.

selection_063

jbalsas: Awesome! Thanks for the details! 👍

PR #10, opened 2019-02-11
Fixes #9 - Remove npm-which package

### LINKED ISSUE #9: Add Yarn Workspaces Support
In a Yarn workspace context, most `liferay-npm-scripts` will fail to run from within the workspaces.

Using the attached simple project:
**From root folder**
- `yarn install`
- `yarn workspace Foo run build` // works!

**From packages/Foo folder**
- `npm run build` // fails!

The main problem is that `liferay-npm-scripts` is the only `devDependency` on the workspace, so yarn will only create that symlink in the `/packages/Foo/node_modules/.bin` folder (as expected). 

The binaries for this package dependencies will end up in `/node_modules/liferay-npm-scripts/node_modules/.bin`, which are unreachable by [npm-which](https://github.com/bryceosterhaus/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/scripts/build.js#L10) that will only look in `CWD` and above.

A possible fix would be to have different `npm-which` resolutions:
```
const cwdWhich = require('npm-which')(process.cwd());
const dirnameWhich = require('npm-which')(__dirname);

let babel;

try {
    babel = cwdWhich('babel');
catch () {
    babel = dirnameWhich('babel');
}
```

Note that scripts will work if ran from the root of the project with the `yarn workspace Foo run build` command.

Since I've only had limited experience with Yarn, I'm not sure what exactly should be expected... @wincent, any thoughts?

[yarn_workspaces.zip](https://github.com/bryceosterhaus/liferay-npm-tools/files/2824565/yarn_workspaces.zip)

bryceosterhaus:


jbalsas:
Seems to be working fine if invoked using `yarn`. I think this can be good enough for our intent to move to Yarn workspaces for now.

@bryceosterhaus, I don't think I have permissions here, could you merge this?

PR #16, opened 2019-02-27
Fixes #15 - Resolves dependencies paths to support hoisted packages in workspace settings

### LINKED ISSUE #15: Improve build (soy) dependencies resolution
We're currently assuming all build dependencies will be available in a local `node_modules` folder. This makes it impossible to work on a workspace setting where modules can be hoisted up in the folder tree.

We could try to make use of node's own module resolution to ensure we can properly work in all scenarios.

jbalsas:


packages/liferay-npm-scripts/src/utils/generate-soy-dependencies.js
```diff
@@ -1,10 +1,29 @@
...
+
+			try {
+				// Requires the `package.json` file to avoid resolving
+				// the main entry point of the package so we can safely
+				// infer the directory from the package root
+				resolvedDependency = path.dirname(
+					require.resolve(`${dependency}/package.json`)
+				);
+			}
+			catch (err) {}
```
matuzalemsteles: hey @jbalsas, I think it would be good to send some error message here in case something out of the ordinary happens, swallowing the error sometimes lets the developer accumulate hours trying to understand where the error is 🙂.

jbalsas: Hey @matuzalemsteles, yeah, I totally agree!

The problem here is that the `dependencies` array we use is merges in https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-liferay.json or https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-metal.json which might or might not be available, so there's very little the developer could do. It would simply add noise to the build when there's nothing wrong with it.

matuzalemsteles: Hmm, but looking at this case: The developer can add a wrong or missing dependency, it would be interesting to help by providing the error saying that not solve the path. So the developer can look again at their dependencies and check it out.

makes sense?

jbalsas: Yes and no :)

The current behaviour is that the compiler, `metal-tools-soy` is forgiving about wrong paths because it uses [`globby`](https://github.com/metal/metal-tools-soy/blob/master/lib/pipelines/compileSoy.js#L82-L84) to find matching files. If it doesn't find one then it just doesn't.

Since the user is providing a glob, things might or might not match, but that should not result in an error. Changing it otherwise would actually introduce a breaking change :)

jbalsas: The developer would know that the dependencies are wrong because the templates will fail to compile... so I think that should provide enough information to infer a fix.

matuzalemsteles: I understood, I think we can go with this. Thanks @jbalsas.

jbalsas: No problem @matuzalemsteles!

Just in case it's not clear, I wholeheartedly agree with you that we should not silently swallow exceptions... just not sure if the alternative is better in this case 😂

matuzalemsteles: Np, I understood your point of view. I had been wondering if we could provide the error sooner than waiting for the `metal-tools-soy` to fail...

It's just not clear to me yet (I have not yet looked at how npm-scripts works), `dependencies` when merged with https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-liferay.json, would the project necessarily be using all or just a few packages?

wincent:
Clever trick, resolving package.json!

PR #22, opened 2019-02-27
Fixes #21 - Scopes soy glob to src folder only

### LINKED ISSUE #21: Adapt build dependencies for a yarn workspace scenario
One of the side-effects of moving towards a yarn workspace setup, is that all declared workspaces are available as packages in the root workspace by their package name.

That means that in our Liferay Portal setup, we can get rid of the `jsCompile` rule since every declared package would be available using Node's module resolution. For this to work, build dependencies should properly match their real names. (`com.liferay.frontend.js.web` => `frontend-js-web`).

In addition, the soy glob selector should be more specific to avoid duplication between the `src` files and the build output (`classes`).

jbalsas:


jbalsas:
I'm going to give this a last attempt for today, so forgive the rude auto-merge 🙏

PR #24, opened 2019-02-28
Fixes #23 - Filters oneself dependency to avoid duplicating templates

### LINKED ISSUE #23: Remove duplicated and invalid soy dependencies
It looks like I was a bit too ambitious with my last so-called [fix for soy dependencies resolution](https://github.com/liferay/liferay-npm-tools/pull/22).

It turns out that:
- Not every portal project is safe to include in `soyDeps`. Some contain incorrect soy templates that will make the build crash. Others contain deprecated versions that will make the build crash. In essence, the build 💥 
- There's some duplication when it comes to oneself compilation. That is, `frontend-taglib` contains soy files... but if it that in turns adds `frontend-taglib` as a `soyDep`, then those templates are found twice. Thus 💥 
    - This was not happening before or it was just being masked by the fact that only in some places where we expanding the `frontend-taglib` artifact. Depending on when that happened (before or after) in relation to one of the default dependencies the build would 💥

jbalsas:


packages/liferay-npm-scripts/src/utils/generate-soy-dependencies.js
```diff
@@ -23,6 +26,7 @@ module.exports = function(dependencies) {
 			return resolvedDependency;
 		})
 		.filter(Boolean)
+		.filter(dependencyPath => dependencyPath !== cwd)
```
jbalsas: I'm adding another pass, just to make it explicit, but we should probably look at ways to optimize this.

jbalsas:
I'm not sure if anyone's awake just yet, so I'm going to go ahead and merge to give the build one final (now for realz!) shot.

I'd be happy to send followups, so don't be shy with the comments :)

PR #31, opened 2019-03-12
Fixes #30 - Prepends current PATH so local binaries have precendence in a workspace environment

### LINKED ISSUE #30: PATH is not expanded for local binaries
When running in a workspace scenario with Yarn, depending on the module resolution, some local versions can conflict with the global ones, specially if the installed one is lower. 

This currently happens with our `metal-soy@4.x` dependency when confronted with a package that uses a higher one `6.x`. 

For example, it can happen that:

```
./node_modules/.bin/metalsoy --> ../../some/other/module/node_modules/.bin/metalsoy (/some/other/module/node_modules/.bin/metalsoy)
./node_modules/liferay-npm-scripts/.bin/metalsoy --> ../../../../metal-tools-soy/bin/index.js (./node_modules/metal-tools-soy/bin/index.js)
```

In the scenario above, when running `spawnSync('metalsoy'...)`, the binary in `some/other/module` will be used found globally in the path rather than the one installed locally.

It seems like we should prepend the current `PATH` in `spawnSync` to allow yarn to locate local binaries.

Other options of course are to make sure we don't run into this and always use the same version of `metal-tools-soy` (tried it, but it's a bit more nuanced than that) or instruct yarn to resolve the top binary to a specific version through resolution (not sure if possible or if will get us into trouble too...)

jbalsas:


packages/liferay-npm-scripts/src/utils/spawnSync.js
```diff
@@ -12,9 +12,9 @@ const {spawn,} = require('cross-spawn');
  * @param {Object=} options={} Options to pass to spawn.sync
  */
 module.exports = function(command, args = [], options = {}) {
-	const env = process.env;
-
-	env.PATH = `${PATH}:${env.PATH}`;
+	const env = Object.assign({}, process.env, {
+		PATH: `${PATH}:${process.env.PATH}`,
+	});
```
wincent: (nit) We're already using spread syntax in this file, so it is probably safe to do this without `Object.assign` like I suggested [here](https://github.com/jbalsas/liferay-npm-tools/commit/6f2ea6cac3208109b29f5db442fd7214eb1273ab#commitcomment-32700044).

Issue #33, opened 2019-03-12
Bundling some clay components breaks the build


andresfulla:
Our `.liferaynpmscriptsrc`

```javascript
{
 "build": {
  "dependencies": [
    "clay-button",
    "clay-checkbox",
    "clay-dropdown",
    "clay-icon",
    "clay-link",
    "clay-list",
    "clay-portal",
    "clay-radio",
    "com.liferay.frontend.js.web",
    "com.liferay.frontend.taglib"
  ]
 }
}
```
The moment I add `clay-list` I get errors complaining about clay components already been declared in Soy.

The when `gradlew deply -a` is ran:
```
/Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-list/node_modules/clay-link/src/ClayLink.soy:87: error: Template 'ClayLink.content' already defined at /Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-link/src/ClayLink.soy:87:1
{template .content}
^

/Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-list/node_modules/clay-link/src/ClayLink.soy:121: error: Template 'ClayLink.icon' already defined at /Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-link/src/ClayLink.soy:121:1
{template .icon}
^

/Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-list/src/ClayList.soy:295: error: Undefined template 'ClayLabel.render'.
                                        {call ClayLabel.render}
     ^

/Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-list/src/ClayList.soy:324: error: Undefined template 'ClaySticker.render'.
                        {call ClaySticker.render}
   ^

/Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-portal/src/ClayPortal.soy:6: error: Template 'ClayPortal.render' already defined at /Users/andresfulla/code/portal/liferay-portal/modules/apps/layout/layout-content-page-editor-web/node_modules/clay-list/node_modules/clay-portal/src/ClayPortal.soy:6:1
{template .render}
^

[...]

PR #34, opened 2019-03-18
Add META-INF path to Soy dependencies

p2kmgcl:
Liferay modules injected with `jsCompile` gradle command do not have a `src` directory, but a `META-INF/...`

/cc @jbalsas 
/cc @boton


boton:
This would fix the #33

jbalsas:
Hey @p2kmgcl, I think we can close this now that we have proper `workspaces` support through yarn in portal, so this shouldn't be needed. What do you think?

p2kmgcl:
@jbalsas true :3

PR #37, opened 2019-03-25
Don't use relative path for jest testResultsProcessor

wincent:
With the move to Yarn workspaces in the portal, this configuration was causing all tests that use liferay-npm-scripts to run their tests to fail, because the reporter module is now hoisted up to the top of the workspace.

It is defined as a dependency of the liferay-npm-scripts, so we can count on it being installed, and just reference it by its name; the node module resolution algorithm will find it.

Associated LPS: https://issues.liferay.com/browse/LPS-92609


jbalsas:
Trivial enough, I had also tested this, so merging.

PR #39, opened 2019-03-25
Change merge order so that local overrides have precedence

wincent:
Noticed while preparing #37 that if I made a local override in my "package.json" it had no effect. That's because liferay-npm-scripts is taking whatever user settings are provided (either in "package.json" or in a "jest.config.js" file) and then overwriting them with the top-level defaults defined in its own "config/jest.json" file.

This seems to be the wrong behavior for defaults, which should only take effect in the absence of a more explicit setting.

Before merging this, we should check the portal repo to make sure we're not relying on the old behavior anywhere; at the very least we can clean out some settings from the repo that can't possibly be having any effect.

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/get-merged-config.js
+++ b/packages/liferay-npm-scripts/src/utils/get-merged-config.js
@@ -16,35 +16,35 @@ module.exports = function(type) {
 		case 'babel':
 			return sortKeys(
 				deepMerge(
-					getUserConfig('.babelrc', 'babel'),
-					require('../config/babel')
+					require('../config/babel'),
+					getUserConfig('.babelrc', 'babel')
 				)
 			);
 			break;
 		case 'bundler':
 			return sortKeys(
 				deepMerge(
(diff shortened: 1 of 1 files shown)


wincent:
Consider this one a bit of an RFC because it could potentially break some things (and with things *already* currently broken a bit, I would be careful before adding this one into the mix too).

jbalsas:
Let's wait on this one for a couple of versions to let everything stabilize!

jbalsas:
Based on some conversations with @carloslancha, I'm leaning on merging this... @wincent, @bryceosterhaus, thoughts?

jbalsas:
Actually, the observation is that user settings usually override completely the defaults, which shouldn't even be the case with this code, so there's something fishy here... 🤔

bryceosterhaus:
This looks like the right fix to me. `deepmerge` will override changes from arg1 that are in arg2 so I think this is correct in that the user config needs to be the _last_ item when merging these objects.

Issue #32, opened 2019-03-12
Apply general formatting guidelines


jbalsas:
While this project is used to enforce consistent build processes like linting and formatting, it itself doesn't follow neither. We should either use `liferay-npm-tools` to lint and format itself or at the very list follow the rest of other `out-of-Portal` projects.

wincent:
Recent examples of current "best practice":

- Use "eslint-config-liferay" ([example](https://github.com/liferay/alloy-editor/blob/934bb406b32125a838fee438cb6f06dccb0bcc5c/package.json#L17))
- Use "prettier" ([example](https://github.com/liferay/alloy-editor/blob/934bb406b32125a838fee438cb6f06dccb0bcc5c/package.json#L43))

Example "package.json" scripts (these ones snipped from alloy-editor, but similar ones in liferay-js-themes-toolkit, liferay-amd-loader etc):

```
"format": "prettier --write -- '*.js' '{scripts,src,test}/**/*.{js,jsx}'",
"format:changed": "git ls-files -mz *.js -- 'scripts/*.js' 'src/*.js' 'src/*.jsx' 'test/*.js' 'test/*.jsx' | xargs -0 prettier --write --",
"format:check": "prettier --list-different -- '*.js' '{scripts,src,test}/**/*.{js,jsx}'",
"lint": "eslint 'src/**/*.{js,jsx}'",
"lint:changed": "git ls-files -mz 'src/*.js' 'src/*.jsx' | xargs -0 eslint",
"lint:fix": "eslint --fix 'src/**/*.{js,jsx}'",
"lint:quiet": "eslint --quiet 'src/**/*.{js,jsx}'",
```

bryceosterhaus:
I went ahead and added `eslint-config-liferay` and using `prettier-eslint`.

wincent:
@bryceosterhaus: One thing I've found in other projects is that `prettier-eslint-cli` has some unexpected behaviors. For example, running `format` ends up being equivalent to `eslint --fix`, which is surprising to say the least (because it can do stuff like adding license headers to files; not something I expect when running prettier). So in other projects, I swapped to `prettier` directly, which doesn't have this problem; examples:

- https://github.com/liferay/liferay-amd-loader/pull/194/commits/30e1dfdf1b516b147d890834fe303b0cbc507fcd
- https://github.com/liferay/liferay-npm-build-tools/pull/290/commits/a11a7c7189291ef192f9e59b7c61e74087bf74ac

[...]

PR #41, opened 2019-03-26
chore: separate formatting and linting scripts

wincent:
Adds these scripts to package.json, in addition to "format" and "format:check" which we already had:

- "lint": reports lints but doesn't fix them.
- "lint:fix": reports lints and fixes autofixable lints.

An earlier version of this commit in the PR had a "fix" script that was equivalent to "format" followed by "lint:fix" but we're temporarily backing that out until we come up with a better name.

Test plan: Run all the new scripts; see "format" fix formatting issues in one file, and "lint" report these issues which I then fixed:

    packages/liferay-npm-scripts/src/index.js
      61:4  error  Unexpected console statement  no-console

    packages/liferay-npm-scripts/src/scripts/eject.js
      36:33  error  Unnecessary escape character: \"  no-useless-escape
      38:5   error  Unnecessary escape character: \"  no-useless-escape

    packages/liferay-npm-scripts/src/utils/generate-soy-dependencies.js
      28:18  error  Empty block statement  no-empty

    packages/liferay-npm-scripts/src/utils/get-merged-config.js
      23:4  error  Unreachable code              no-unreachable
      31:4  error  Unreachable code              no-unreachable
      39:4  error  Unreachable code              no-unreachable
      50:4  error  Unreachable code              no-unreachable
      52:4  error  Unexpected console statement  no-console

    ✖ 9 problems (9 errors, 0 warnings)

Closes: https://github.com/liferay/liferay-npm-tools/issues/32

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,10 @@
 	},
 	"private": true,
 	"scripts": {
-		"format": "prettier --write \"./**/*.js\"",
-		"format:check": "prettier --list-different \"./**/*.js\""
+		"format": "prettier --write '**/*.js'",
+		"format:check": "prettier --list-different '**/*.js'",
+		"lint": "eslint '**/*.js'",
+		"lint:fix": "eslint --fix '**/*.js'"
 	},
 	"workspaces": [
 		"packages/*"
--- a/copyright.js
+++ b/copyright.js
@@ -3,3 +3,4 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+
--- a/packages/liferay-jest-junit-reporter/__tests__/index.js
+++ b/packages/liferay-jest-junit-reporter/__tests__/index.js
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+
 jest.mock('fs');
 
 const reporter = require('../index');
--- a/packages/liferay-jest-junit-reporter/index.js
+++ b/packages/liferay-jest-junit-reporter/index.js
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+
 'use strict';
 
 const xml = require('xml');
--- a/packages/liferay-npm-scripts/bin/liferay-npm-scripts.js
+++ b/packages/liferay-npm-scripts/bin/liferay-npm-scripts.js
@@ -1,7 +1,9 @@
 #!/usr/bin/env node
+
 /**
  * © 2019 Liferay, Inc. <https://liferay.com>
  *
  * SPDX-License-Identifier: BSD-3-Clause
  */
+
 require('../src/index')();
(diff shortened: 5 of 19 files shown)


package.json
```diff
@@ -3,10 +3,14 @@
...
 	},
 	"scripts": {
-		"format": "prettier-eslint \"./**/*.js\" --write",
-		"lerna": "npm install && lerna bootstrap"
+		"fix": "npm run format && npm run lint:fix",
+		"format:check": "prettier --list-different '**/*.js'",
+		"format": "prettier --write '**/*.js'",
+		"lerna": "npm install && lerna bootstrap",
+		"lint": "eslint '**/*.js'",
+		"lint:fix": "eslint --fix '**/*.js'"
```
bryceosterhaus: How do you feel about changing the names up a bit? Using `format`, `fix`, and `lint:fix` are all a bit confusing without looking at what they actually do, especially `format` and `fix`.

I'm thinking something like 
```json
{
	"format": "npm run prettier && npm run lint:fix",
	"prettier:check": "prettier --list-different '**/*.js'",
	"prettier": "prettier --write '**/*.js'",
	"lerna": "npm install && lerna bootstrap",
	"lint": "eslint '**/*.js'",
	"lint:fix": "eslint --fix '**/*.js'"
}
```

wincent: I'm open to changing the names but I can give you some insight into why they are named they are:

- "format" and "format:check" instead of "prettier" and "prettier:check" in order to say what they do instead of how they do it (an implementation detail); ie. the same reason that we have "test" script that runs `jest` instead of creating a "jest" one. I know it's unlikely we'll ever use anything but Prettier for formatting, but certainly in the case of testing and linting we have legacy projects that use other tools (Karma, Mocha etc).
- "lint" and "lint:fix" for similar reasons.

So I think those are good. "fix", however, is the worst name: I only added it because you wanted one command that did both linting and fixing. I don't think "format" is a good name for that task, because the whole point of this change is to decoupling the concerns of formatting and linting. Do you have any other suggestions for what a good name for this "fix all the things task"?

wincent:
#43 will conflict with this, so I'll rebase/change/update this after that goes in.

package.json
bryceosterhaus: Ah I see, I had figured you wanted to go the route of "say what they do instead of how they do it", which I am good with, I think it was the "fix" one that through me off when it is coupled with the rest.

Since the "fix" script was added for my concern of doing it all in one command, I'd say just go ahead and remove it. I can just manually run those myself with `npm run format && npm run lint:fix`.

TLDR: Remove fix script, keep the rest as is.

wincent: > TLDR: Remove fix script, keep the rest as is.

Disappointed that you don't like the name, "do:not:use:or:you:will:be:fired". 😢

bryceosterhaus: 😂

First I thought, maybe call it "fix:all", but then I realized I just don't really need it and I'm always a big fan of removing code rather than adding

wincent:
Will shipping this so that I can do the branch restructuring described in #47.

PR #43, opened 2019-03-27
Remove Lerna, add Yarn workspaces, update Jest and Babel

wincent:
Closes: https://github.com/liferay/liferay-npm-tools/issues/42

Issue #42, opened 2019-03-27
feat: update Yarn, Babel, Jest dependencies


wincent:
- [x] Remove basically unused Lerna set-up.
- [x] Switch to Yarn workspaces.
- [x] Update Jest and Babel dependencies.

bryceosterhaus:
@wincent So does yarn workspaces essentially remove the need of lerna? I had always thought it was weird to have both coexist but I never took the time to look into it.

wincent:
Not exactly because there is some overlap between the tools.

Yarn workspaces take care of the linking between packages, and arguably does it in a much more reliable way than Lerna because it is baked right in to `yarn install`.

The thing that Lerna still does and which Yarn does not is provide a way to release all the packages at once. Having said that, in this repo, with only three packages, and in all likelihood, only one of them that we are going to release with any regularity ("liferay-npm-scripts"), that is not much of a value-add. The cost is a relatively large dependency footprint, which may or may not be worth it depending on your point of view. And I find it pretty disappointing that Lerna has been so slow to make publishing with 2FA livable; in the end I think you can get almost all the value that Lerna provides with a couple of small scripts.

PR #46, opened 2019-03-27
chore: Update dependencies on liferay-npm-build-tools packages (#42)

wincent:
Ideally this would have been done in #43 as part of the solution to #42, but I didn't realize we needed to update these packages as well.

You know the change is good when it only leads to red lines in the lockfile diff.


wincent:
Seems trivial, so merging.

Issue #38, opened 2019-03-25
Handle spawn errors in utils/spawnSync.js


izaera:
We are not checking for errors when executing tools like babel or the bundler so, if for any reason the tools are not available, the build continues without any failure or message at all thus creating suffering among living beings.

It would be good if we stopped the build when such errors are encountered.

PR #56, opened 2019-04-01
fix: don't silently swallow failures in spawnSync (#38)

wincent:
Test plan: `yarn test` in "liferay-npm-scripts". By way of an integration test, also do this (which will error out due to a config clash):

```
../../node_modules/.bin/liferay-npm-scripts test
● Validation Error:

  Configuration options testMatch and testRegex cannot be used together.

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html

/Users/greghurrell/code/liferay-npm-tools/packages/liferay-npm-scripts/src/utils/spawnSync.js:36
                throw new Error(
                ^

Error: Command jest --config /Users/greghurrell/code/liferay-npm-tools/packages/liferay-npm-scripts/TEMP_jest.config.json exited with code 1
zsh: exit 1     ../../node_modules/.bin/liferay-npm-scripts test
```

Closes: https://github.com/liferay/liferay-npm-tools/issues/38

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/test.js
+++ b/packages/liferay-npm-scripts/src/scripts/test.js
@@ -23,7 +23,7 @@ const JEST_CONFIG = getMergedConfig('jest');
 /**
  * Main script that runs `jest` with a merged config
  */
-module.exports = function(arrArgs) {
+module.exports = function(arrArgs = []) {
 	const useSoy = soyExists();
 
 	const CONFIG_PATH = path.join(CWD, 'TEMP_jest.config.json');
--- a/packages/liferay-npm-scripts/src/utils/spawnSync.js
+++ b/packages/liferay-npm-scripts/src/utils/spawnSync.js
@@ -11,6 +11,10 @@ const PATH = path.resolve(`${__dirname}/../../node_modules/.bin`);
 
 const {spawn} = require('cross-spawn');
 
+function getDescription(command, args) {
+	return [command].concat(args).join(' ');
+}
+
 /**
  * Wrapper function for spawning a synchronous process.
  * @param {string} command Path to bin file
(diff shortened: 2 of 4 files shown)


packages/liferay-npm-scripts/src/scripts/test.js
```diff
@@ -23,7 +23,7 @@ const JEST_CONFIG = getMergedConfig('jest');
 /**
  * Main script that runs `jest` with a merged config
  */
-module.exports = function(arrArgs) {
+module.exports = function(arrArgs = []) {
```
wincent: Note that I added this to make manual QA a bit easier. I noticed that I couldn't just do this in the console due to the expected array argument:

```
require('./src/scripts/test')()
```

packages/liferay-npm-scripts/src/utils/spawnSync.js
```diff
@@ -27,4 +31,20 @@ module.exports = function(command, args = [], options = {}) {
 		stdio: 'inherit',
 		...options
 	});
+
+	if (status) {
+		throw new Error(
```
jbalsas: In the case of `status`, I was wondering... are we piping the spawned process stderr and stdout to the main one? Otherwise, we might be missing not just the exit code, but valuable information.

izaera:
Who does the exit(1)? The standard unhandled exception handler?

packages/liferay-npm-scripts/src/utils/spawnSync.js
izaera: It says `stdio: 'inherit'` so it should be output to the console.

jbalsas: So, from this, we should be seeing the errors in the build console, but just not signaling it properly? Just wondering about the EMF issue you're running into...

izaera: Yes. The EMFILE is seen in the console, in fact. The problem is that because it doesn't stop the build, nobody notices it.

wincent: Just as an example, if you look in my test plan above you can see what happened when I ran `../../node_modules/.bin/liferay-npm-scripts test` in the console with a conflicting jest config (you could see the error output on the console).

wincent:
> Who does the exit(1)? The standard unhandled exception handler?

@izaera: that's right.

PR #61, opened 2019-04-02
Fixes #55 - Add ability to override default config

### LINKED ISSUE #55: I'd like to configure which globs get sent to CSF
CSF complains about not being able to find css and soy files, which makes sense because I don't have any, but I can't prevent `liferay-npm-scripts` from passing those globs to csf.

Example errors:
```
File does not exist: /Users/drewbrokke/Documents/liferay/liferay-portal/modules/apps/users-admin/users-admin-web/src/**/*.css

File does not exist: /Users/drewbrokke/Documents/liferay/liferay-portal/modules/apps/users-admin/users-admin-web/src/**/*.soy
```

bryceosterhaus:
Hey @jbalsas this is a general idea I had for what we were talking about in regards to overriding/merging configs.

Rather than add another field to the existing config and rather than changing the behavior for the way we merge other types of config(babel, jest, etc). I went with the idea of the user creating another rc file that allows configuration to override rather than merge. We won't have to make a breaking change with this and it will also allow us developers to use something in the mean time while we try and iron out the higher level ideas behind the tooling. I also made sure to note in the README that this solution is not encouraged and also likely will go away in the future, but I think it provides us some flexibility in the mean time.

Let me know what you think.

### DIFF
--- a/packages/liferay-npm-scripts/README.md
+++ b/packages/liferay-npm-scripts/README.md
@@ -77,6 +77,38 @@ Eject will remove `liferay-npm-scripts` as a dependency and write all of the nec
 
 If you need to add additional configuration you can do so by creating a `.liferaynpmscriptsrc` file at the root of your project. The default configuration of this file can be seen [here](./src/config/liferay-npm-scripts.json).
 
+#### Overriding default arrays
+
+In the rare case you want to override default config such as the default dependency list or lint/format globs, we provide an escape hatch via `.liferaynpmscriptsrc-override`. By using this file, you are able to override a configuration rather than merging with the default values. See examples below for more info.
+
+The example below would result in _only_ `foo-bar` being passed as a soy dependency.
+
+```json
+{
+	"build": {
+		"dependencies": ["foo-bar"]
+	}
+}
+```
+
+The example below would result in _only_ jsx files being linted, instead of adding `**/*.jsx` to the pre-existing [defaults](./src/config/liferay-npm-scripts.json#L14-L20).
+
+```json
+{
+	"lint": ["**/*.jsx"]
+}
+```
+
+The example below would result in passing no globs to the lint task and override the [default globs](./src/config/liferay-npm-scripts.json#L14-L20).
+
+```json
+{
+	"lint": []
+}
+```
+
+_Using `.liferaynpmscriptsrc-override` is not recommended and will likely be removed in the next major version change._
+
 ### Other Config
 
 If you need more flexibility over babel or the bundler. You can still add a `.babelrc` or `.npmbundlerrc` which will be merged with the default settings this tool provides. [Default Babel Config](./src/config/babel.json), [Default Bundler Config](./src/config/npm-bundler.json)
--- a/packages/liferay-npm-scripts/src/utils/deep-merge.js
+++ b/packages/liferay-npm-scripts/src/utils/deep-merge.js
@@ -31,11 +31,24 @@ function combineMerge(target, source, options) {
 }
 
 /**
+ * Code copied from https://github.com/TehShrike/deepmerge#combine-array
+ */
+const overwriteMerge = (destinationArray, sourceArray) => sourceArray;
+
+/**
  * Helper to get merge two json objects
  * @param {Object} defaultConfig Config file
  * @param {Object} customConfig Config file
  * @returns {Object}
  */
-module.exports = function() {
-	return merge.all([...arguments], {arrayMerge: combineMerge});
+exports.deepMerge = function() {
+	return merge.all([...arguments], {
+		arrayMerge: combineMerge
+	});
+};
+
+exports.deepMergeOverWrite = function() {
+	return merge.all([...arguments], {
+		arrayMerge: overwriteMerge
+	});
 };
--- a/packages/liferay-npm-scripts/src/utils/get-merged-config.js
+++ b/packages/liferay-npm-scripts/src/utils/get-merged-config.js
@@ -6,7 +6,7 @@
 
 const sortKeys = require('sort-keys');
 const getUserConfig = require('./get-user-config');
-const deepMerge = require('./deep-merge');
+const {deepMerge, deepMergeOverWrite} = require('./deep-merge');
 
 /**
  * Helper to get JSON configs
@@ -40,12 +40,15 @@ module.exports = function(type) {
 
 		case 'npmscripts':
 			return sortKeys(
-				deepMerge(
-					require('../config/liferay-npm-scripts'),
-					require('../config/liferay-npm-scripts-build-deps-clay.json'),
-					require('../config/liferay-npm-scripts-build-deps-liferay.json'),
-					require('../config/liferay-npm-scripts-build-deps-metal.json'),
-					getUserConfig('.liferaynpmscriptsrc')
+				deepMergeOverWrite(
+					deepMerge(
+						require('../config/liferay-npm-scripts'),
+						require('../config/liferay-npm-scripts-build-deps-clay.json'),
+						require('../config/liferay-npm-scripts-build-deps-liferay.json'),
+						require('../config/liferay-npm-scripts-build-deps-metal.json'),
+						getUserConfig('.liferaynpmscriptsrc')
+					),
+					getUserConfig('.liferaynpmscriptsrc-override')
 				)
 			);
 


wincent:
I'd be wary about adding something with imminent plans to remove it (removing is much harder than adding, once people depend on something; and especially if you document it in the readme, people are more likely to use it — if you do merge this, I'd suggest hiding the "documentation" in a comment in the source code, so that it is not so much an "escape hatch" as a "secret escape hatch").

Just to mention a few other possible ways of addressing #55:

- One would be to simply not issue warnings when globs don't match.
- Another would be to make the warning less noisy; for example:
  - We could print relative paths:
    - `File does not exist: modules/apps/users-admin/users-admin-web/src/**/*.css`; or even:
    - `File does not exist: users-admin-web/src/**/*.css` (or possibly less); instead of:
    - `File does not exist: /Users/drewbrokke/Documents/liferay/liferay-portal/modules/apps/users-admin/users-admin-web/src/**/*.css`
  - We could make the warning text less scary:
    - `Skipping glob with no matches: ...`
    - `File does not exist: ...`
  - We could print the glob pattern instead of the path:
    - `Skipping glob with no matches: src/**/*.css`; instead of:
    - `File does not exist: modules/apps/users-admin/users-admin-web/src/**/*.css`
  - We could eliminate the blank lines that appear between the warnings, and which make them take up a lot of space.

If we really, really want to implement overrides, unfortunately we don't control all of the config file formats (for example, not the Babel and Jest ones), but I think the "right" way to do this kind of pattern (composing and overriding configs in a hierarchical context) in the general sense is to let people use explicit "extends" and "overrides" directives. Projects like Babel and ESLint offer both an "extends" and an "overrides" config directive and document exactly how they work. It is not perfect, and it is pretty complicated, but it does allow you to implement both "merge" and "replace" semantics. Not saying we should to that here — keeping the complexity and variations low by being opinionated and having strong defaults is probably a good starting policy — but in the abstract/general sense, "extends"/"overrides" is probably the right approach.

matuzalemsteles:
Hey guys, just also leaving my thoughts on this, I agree with @wincent. I think maybe we do not necessarily have to just hide that functionality, we can treat it as `unstable` or `experimental` may give us a bit of flexibility to remove if it is the wrong way we are going and issue a warning when someone uses this API `experimental` .

bryceosterhaus:
Just to give a little more background on the issue, since its not entirely all because of issue #55.

We run into issues with the pre-configured dependencies when you are trying to use `liferay-npm-scripts` in a module within that last. For example, if we wanted to use the scripts in `frontend-taglib`, we would run into issues because we define this as a dependency [here](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-liferay.json#L7). So we need to provide a way for users to override for these cases. So this isn't just a fix for our csf output.

Now, I agree that this is not an ideal solution of creating an `-override` file. However, I think its a quick solution we can provide now as we iterate and aim for a better solution in the next major change. I also like this solution(for now) since it only deals with the `.liferaynpmscriptsrc` config, and it doesn't effect our jest or babel configs. So I do think this is a good _intermediate_ solution. For the documentation, I like the idea of hiding it a bit, how about we just add some comments here, [here](https://github.com/liferay/liferay-npm-tools/pull/61/files#diff-0824ae1ca79bc9d7f1cdc1a9c5c6f159R43). And again, the primary reason I think we need functionality sooner rather than later is so that we can use it internally for modules like `frontend-taglib` and override the build dependencies.

As for a long term solution, I like this idea of allowing users to "extend", that way it can be explicit to users what they are actually using. I think this needs to be investigated more heavily before doing so though.

jbalsas:
Hey @bryceosterhaus, I'm still hesitant to allow this type of escape hatch in our DXP builds, specially after what we're seeing and how hard it is to normalize projects back once they've been able to deviate from the common path.

In the end, it looks like we're giving developers an out so they don't get blocked, but at the same time, we're letting them roam free and without support. Down the road, they end up with more questions they can answer on things not even ourselves can help...

I think we can keep this open a couple more days until we finalize the migration to `npm-scripts` and probably start applying the formatting rules to see if this is really necessary or we can push what's missing through the same tube...

I'll work with @diegonvs on migrating the last remnants of custom builds and see how it looks from there.

Thanks!! :)

jbalsas:
I gave this a little more thought and it came to me that by being more strict we might be able to better channel the conversation and circle back to that idea that we wanted to explore of the "standardizing" working groups.

I've been trying to come up with a good way to orchestrate that in a way that is fruitful, open and avoids unnecessary frustration by lack of context and impossible moonshots...

**If** we were to be more restrictive, new additions and practices would be discussed right over here and could be easily rolled out and communicated throughout our developers. Imagine some team wants to start using decorators (stage2). Since it's not enabled by default and not allowed by the tool, the team would need to open a conversation over here and explain the merits, pros and cons of the approach and the best way to roll it out could be agreed together...

Just an additional thought I wanted to share :)

bryceosterhaus:
@jbalsas I actually really like that idea. We've never really had a single source to come to in regards to what things we allow to be used in portal. By having this, it provides us with a unique opportunity to be restrictive *so that it leads to discussion* rather than someone just adding something or modifying their own process for their module.

However, I think it might be wise for us to isolate that to a preset of some sort. That way the tool stays generic(flexibility if needed for customers), but then the preset that is actually used in portal would be locked down and restricted. That way when a liferay dev asks what they can use, we point them to the preset and that can be the source of truth for what is allowed.

What do you think?

matuzalemsteles:
Just leaving my thoughts on this: I like the idea of ​​creating working groups for standardization, this is much more democratic and we can have the vision and contribution of all, this is very enriching.

I am doing something similar to Brazil's Frontend devs, something I call TC9 (Technical Committee and 9 are the amount of Frontend we currently have 🙂), basically the idea is to hold meetings only when necessary to talk about some topic and we can discuss which actions we should take but the meeting follows a [Framework called DACI](https://www.atlassian.com/team-playbook/plays/daci) to help make the decision and the people invoking the meeting have to fill a template to give context to such, the meeting output is what I call the **Pattern Decision Records - PDR** (Inspired by the [ADR's](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions) process), we keep these decisions documented so that new people can see. I introduced this a few days ago so I have no clear feedback on how it's going and is working out. Maybe soon...

I do not like the idea of ​​having to create something of the very bureaucratic type, which can take a few days to be solved and many discourses and possibly will block some team due to the delay of response.

I'd rather have an exclusive repo created to store these decisions, rejected, approved and in progress so people can send PR's according to the template (something like RFC..., ADR), so we'll discuss and decide if it's going to be approved as standard or not.

I also like the idea of ​​@bryceosterhaus to leave it flexible and add a preset that sets the defaults and this preset is used inside Liferay, but I think it would make sense to do this only if we will make this tool available for use by customers or something.

wincent:
> We run into issues with the pre-configured dependencies when you are trying to use `liferay-npm-scripts` in a module within that last. For example, if we wanted to use the scripts in `frontend-taglib`, we would run into issues because we define this as a dependency here. So we need to provide a way for users to override for these cases. So this isn't just a fix for our csf output.

About a solution for this, I get the impression that we could deal with this internally (ie. with some kind of graceful circular-dependency check, or an internal black list), without having to expose an "escape hatch", which as @jbalsas said, may actually end up being a "wide-open barn door" (well, he didn't actually say that, but I like to think that he would). This would involve some internal complexity (bad), but wouldn't create an external liability (also bad); the sense I get — an emerging consensus, maybe? — is that the latter is more harmful and the one we should try harder to avoid.

bryceosterhaus:
I'd be fine with adding internal complexity to get around the issue. We'd just need to make sure that it doesnt skip over _all_ circular dependencies. Its be helpful for devs to see errors if they accidentally list multiple of the same dependency or conflicting dependencies.

I still think presets for the default configs is ideal though. I think of something like babel's plugin/preset system and its always been a super helpful paradigm for me when working with it because its so easy to follow where all of the presets come from and what they are doing. I like the explicit nature of it, rather than what we have here, where a user will inherit a bunch of default dependencies that they may not even use or be aware of.

jbalsas:
Hey @bryceosterhaus, `presets` have worked fine for `npm-bundler` in this case. We have our `standard` preset (which is mostly empty and portal-unaware) which we can use to bundle the infra modules and the `liferay-dev` one that we use for all the others.

I'd be fine if we take as similar approach for the other bits of the tools, so we can build from an almost empty preset for the infra modules and still use by default a more liferay-dxp-aware one for all other ones so we can achieve an almost 0-config setup inside `liferay-portal`.

What do you think?

bryceosterhaus:
Yeah that sounds good to me. So just to confirm, the idea going forward would be...

- Allow `liferay-npm-scripts` to accept "presets" for the build config
- A "preset" is composed of a list of build dependencies that essentially get merged [here](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts.json#L3)
- `liferay-npm-scripts` by itself should have a dependencies list of `[]`.
- Any non-infra dxp modules should use some sort of "dxp-preset" that is comprised of [clay](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-clay.json), [liferay](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-metal.json), and [metal](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/liferay-npm-scripts-build-deps-liferay.json)
- Infra modules should rely on a preset that is more of a subset of the `dxp-preset`

And then additionally maybe it should...
- Default to use the `dxp-preset` but add ability to override presets.


This would give us nearly a 0-config setup for non-infra modules, and then for infra specific modules we would just override the preset with one of our own.

That seem accurate? Once we narrow this down, maybe we can copy this info over to an issue and then I can look into adding it later this week.

jbalsas:
Going forward, if we take on the `preset` approach, I'd say we can get rid of the `merge` step and replace it with the `preset` extension mechanism. 

For example, `npm-scripts` should **not** have an empty list of dependencies, but a default `preset` that is very dumb and lean. That would be the preset we use to build infra modules (for bundler, we have the [`standard` preset](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-taglib/frontend-taglib-clay/.npmbundlerrc#L38) and the [`liferay-dev` preset](https://github.com/liferay/liferay-npm-tools/tree/master/packages/liferay-npm-bundler-preset-liferay-dev)

This is to say, we should not need to add ability to override presets, but to provide simpler ones that can be used. That way, we just augment and remove the need of overriding presets.

Makes sense?

bryceosterhaus:
Not quite sure what you mean by 
> This is to say, we should not need to add ability to override presets, but to provide simpler ones that can be used. That way, we just augment and remove the need of overriding presets.

How would this look practically?

Additionally, if we remove the merge step for build dependencies, how should we also extrapolate this for other configs such as the lint and format globs?

bryceosterhaus:
new pr https://github.com/liferay/liferay-npm-tools/pull/76

Issue #64, opened 2019-04-04
Add support for running theme builds


wincent:
If you look at https://github.com/brianchandotcom/liferay-portal/pull/70792/commits/5fa16c09e89df8bac91e9f8b708f382ffe3c11ac you'll see that we are passing a long `gulp build` invocation as a package.json `"build"` script for the themes in liferay-portal. This is sufficiently complicated that we'd like to hide away the details in liferay-npm-scripts: absolutely everything in the portal should be depending on the same (ie. local, "master") version of the styled and unstyled theme anyway.

PR #65, opened 2019-04-04
feat: add "theme" subcommand (#64)

wincent:
With this, we can replace scripts like theses ones in liferay-portal:

    "build": "gulp build --css-common-path ./build_gradle/frontend-css-common --styled-path ../../frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled --unstyled-path ../../frontend-theme/frontend-theme-unstyled/src/main/resources/META-INF/resources/_unstyled"

with:

    "build": "liferay-npm-scripts theme build"

I added a bunch of tests because this involves traversing the filesystem and I didn't want to screw it up. Using runtime traversal rather than hardcoding relative paths in order to make us _somewhat_ resilient to possible future changes in the structure of the filesystem layout. As it is right now, we have two variants in liferay-portal, which I have reflected in the structure of the fixtures:

- Marketplace themes like fjord that live in modules like "frontend-theme-fjord/frontend-theme-fjord".
- Core themes which like "classic" that live under "frontend-theme",  right next to the "unstyled" and "styled" dependencies.

Test plan: `yarn test`

Closes: https://github.com/liferay/liferay-npm-tools/issues/64

### DIFF
--- a/packages/liferay-npm-scripts/__fixtures__/scripts/theme/modules/apps/frontend-theme-fjord/frontend-theme-fjord/.gitignore
+++ b/packages/liferay-npm-scripts/__fixtures__/scripts/theme/modules/apps/frontend-theme-fjord/frontend-theme-fjord/.gitignore
@@ -1,0 +1,2 @@
+*
+!/.gitignore
--- a/packages/liferay-npm-scripts/__fixtures__/scripts/theme/modules/apps/frontend-theme/frontend-theme-classic/.gitignore
+++ b/packages/liferay-npm-scripts/__fixtures__/scripts/theme/modules/apps/frontend-theme/frontend-theme-classic/.gitignore
@@ -1,0 +1,2 @@
+*
+!/.gitignore
--- a/packages/liferay-npm-scripts/__fixtures__/scripts/theme/not/the/apps/you/are/looking/for/.gitignore
+++ b/packages/liferay-npm-scripts/__fixtures__/scripts/theme/not/the/apps/you/are/looking/for/.gitignore
@@ -1,0 +1,2 @@
+*
+!/.gitignore
--- a/packages/liferay-npm-scripts/__tests__/scripts/theme.js
+++ b/packages/liferay-npm-scripts/__tests__/scripts/theme.js
@@ -22,7 +22,7 @@ const UNSTYLED = path.join(
 	'frontend-theme-unstyled/src/main/resources/META-INF/resources/_unstyled'
 );
 
-describe('', () => {
+describe('scripts/theme.js', () => {
 	let cwd;
 	let spawnSync;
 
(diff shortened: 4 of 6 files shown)


jbalsas:
LGTM, let's see what @bryceosterhaus has to say about it! :)

bryceosterhaus:
Good idea. Only concern I have would be naming, rather than `liferay-npm-scripts theme build` can we get away with just `liferay-npm-scripts theme`? Otherwise, do we start duplicating lint and format for themes specifically as well and do we rename build to `liferay-npm-scripts portlet build` to keep consistent? 

Or even possibly, we could do `liferay-npm-scripts build --theme` since we have that pattern with `--soy`.

Overall, its not a huge deal to me, just thinking out loud how we can be consistent with naming. If this is the best we got, I'd still be fine with it.

jbalsas:
Hey @bryceosterhaus, we had a similar conversation offline... iirc, @wincent comment was _everything is a verb, so `scripts theme` wouldn't match...`.

> Or even possibly, we could do liferay-npm-scripts build --theme since we have that pattern with --soy.

This is something I've been meaning to bring up, because I think this is no longer really the case... I don't remember when, but I think we made `build` implicit (smart?). Correct me if I'm wrong, but I think that simply `liferay-npm-scripts build` would [runSoy if soy files exist](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/scripts/build.js#L66), [always run babel and bundler](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/scripts/build.js#L72-L74) and [runBridge if `.npmbridgerc` exists](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/scripts/build.js#L76).

I'm not sure if all this "simplicity" is good or bad, but if we were to follow, we could do `liferay-npm-scripts build` build the theme if `liferayTheme` entry is found in the `package.json`, for example... 🤔

bryceosterhaus:
Oh good call, I totally forgot I made those flags implicit... 🤦‍♂️ 

I'd be all for including the theme to be picked up automatically. So long as developers know that its doing some sort of "build" I think it should be fine to be fairly implicit. Plus that gives us flexibility behind the scenes to add or remove certain functionality or tooling without having to provide a new API.

wincent:
One thing about the naming here is that `liferay-npm-scripts theme` is just a wrapper for the gulp tasks in "liferay-theme-tasks". So, we happen to be running the "build" task when you do `liferay-npm-scripts theme build` (adding the custom options along the way), but you could also run any theme task with it (eg. `liferay-npm-scripts theme watch`, `liferay-npm-scripts theme status` etc). Now, I don't know if you'd ever want to run them in liferay-portal, but you can.

jbalsas:
That's a good point, @wincent!

Since we're the main consumers of this and @bryceosterhaus didn't super-object... I think we could simply merge and release as is and see where this endeavour takes us!

wdt, @wincent?

wincent:
Sounds fine to me @jbalsas.

bryceosterhaus:
👍

PR #68, opened 2019-04-05
feat: add top-level shortcut to run all tests (#67)

wincent:
This it the simplest thing I could think of to get tests running from the top-level. Ideally though, we could just run jest once at the top. That would be pretty easy to set-up, but would require a bit of extra work to make it happen while still preserving the ability to run the tests at the level of the individual packages, so I am keeping it simple for now.

Test plan: `yarn test` at top level and see all run. Then `yarn test` at individual package level to prove that that still works too.

Closes: https://github.com/liferay/liferay-npm-tools/issues/67

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,8 +12,7 @@ git pull --ff-only upstream master
 yarn format:check
 
 # Run tests (currently, we don't have many of them):
-(cd packages/liferay-jest-junit-reporter && yarn test)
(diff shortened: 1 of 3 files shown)


jbalsas:
LGTM, @bryceosterhaus, I'm merging away, but please, do comment when you see this!

jbalsas:
FYI, I've also asked to get Travis-CI enabled for this repo so we can get them through CI

bryceosterhaus:
🥇

Issue #72, opened 2019-04-10
Teach liferay-npm-scripts to look for Sass dependency in top-level node_modules


wincent:
The idea is to make use of the new `--sass-include-paths` setting (which will be added to the liferay-theme-tasks in https://github.com/liferay/liferay-js-themes-toolkit/pull/306) so that we can get themes to look up Sass dependencies in the top-level "node_modules" folder in the liferay-portal repo.

This is necessitated because of changes to where/how we use font-awesome in the portal.

Internal ticket: https://issues.liferay.com/browse/LPS-93751

PR #73, opened 2019-04-10
feat: look for Sass dependency in top-level node_modules (#72)

wincent:
The idea is to make use of the new `--sass-include-paths` setting which will be added to the liferay-theme-tasks in https://github.com/liferay/liferay-js-themes-toolkit/pull/306 so that we can get themes to look up Sass dependencies in the top-level "node_modules" folder in the liferay-portal repo.

This is necessitated because of changes to where/how we use font-awesome in the portal.

Closes: https://github.com/liferay/liferay-npm-tools/issues/72


packages/liferay-npm-scripts/__tests__/scripts/theme.js
```diff
@@ -8,10 +8,12 @@ const os = require('os');
 const path = require('path');
 
 const FIXTURES = path.resolve(__dirname, '../../__fixtures__/scripts/theme');
-const APPS = path.join(FIXTURES, 'modules/apps');
+const MODULES = path.join(FIXTURES, 'modules');
```
wincent: Note that this commit is not the smallest possible edit that I could have made, but I figured it makes sense to treat "modules" as the root instead of "modules/apps" now that we're dealing with something at that level (ie. "modules/node_modules").

Issue #50, opened 2019-03-27
Add yarn "preversion" and "postversion" scripts to semi-automate publishing


wincent:
Based on discussion [over here](https://github.com/liferay/liferay-npm-tools/pull/49#discussion_r269610927) we might consider adding [some "postversion" scripts](https://yarnpkg.com/lang/en/docs/cli/version/) to make cutting releases a little bit easier.

They could be something like:

```json
{
  "scripts": {
    "preversion": "yarn format:check && yarn test",
    "postversion": "git push upstream master --follow-tags && yarn publish"
  }
}
```

or something like that. You'd then publish with `yarn version --minor` etc. I am a bit scared of running two publish operations (a `git push` and a `yarn publish`) like that, but perhaps with some extra checks (like ensuring you're on the "stable" branch etc) would make it feel a bit safer.

In this repo where we need tags of the form "package-name/version" we might need to add a layer of wrapping to ensure the correct format. Or perhaps `yarn config set version-tag-prefix "package-name/v"` and `yarn config set version-git-message "package-name/v%s"` would suffice.

bryceosterhaus:
One issue I see with this is hardcoding the `upstream` remote. I sometimes like to use `github` at my remote name. Might not be a major issue, but I could see it coming up.

> I am a bit scared of running two publish operations (a git push and a yarn publish)
We also should probably check correct permissions to both the git repo and npm package itself. I know in the past I have run into issues when combining both if I am not logged in to the CLI or incorrect permissions.

One other possibility to help automate some of this would be to use Github actions. I've seen other repos do this, although I'm not sure about pain points with it yet since GH actions are relatively new.

wincent:
Yeah, I have the same thought about `upstream`. For example, in most of my checkouts, I actually have `upstream` which uses an HTTPS URL and `upstream-rw` (read-write) that uses an SSH URL, to make it harder for me to accidentally push directly. And then there will be random clones where it ends up being `origin` or `github`. 🤦‍♂️

I think the way to make this agnostic would actually be to put in a wrapper script that inspected the output of `git remote -v` and if it can't find the remote it is looking for complains and tells you to set one up.

PR #74, opened 2019-04-11
feat: simplify release process (#50)

wincent:
As you can see in the changes to CONTRIBUTING.md, the release process can be made simpler by doing two things:

- Adding `.yarnrc` config files that set up the templates for the correct tag and release commit message formats.
- Using `"preversion"` scripts to automatically run the CI checks (formatting, linting, and testing) before cutting a release.

Note that we run the checks for *all* packages in the repo, which is overkill, but it is the simplest thing to do. If a problem in package A ever blocks you from publishing package B you can pass `--ignore-scripts`, but you should probably just fix package A because we want the CI to be green on every commit anyway. As a follow-up, we should make it possible to run the tests independently within a given repo.

There are still further steps we can take to make releasing even simpler (see the related issue), but this is a good first step.

Test plan:

Prepare some pretend releases using different invocations of `yarn version`, see the CI runs succeed, and then inspect output of `git log -p` to see that created commits include the right messages and tags.

    cd packages
    cd liferay-jest-junit-reporter
    yarn version --minor
    cd ../liferay-npm-scripts
    yarn version --patch
    cd ../liferay-npm-bundler-preset-liferay-dev
    yarn version
    git log -p
    git tag -l

Related: https://github.com/liferay/liferay-npm-tools/issues/50

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,21 +8,15 @@ At the time of writing we have a pretty informal release process for the package
 git checkout master
 git pull --ff-only upstream master
 
-# Make sure formatting is up-to-date:
-yarn format:check
-
-# Run tests (currently, we don't have many of them):
-yarn test
-
 # Update individual package versions:
 # - Note that this time we updated all the packages,
 #   but on many occasions we'll update only one, or two.
 cd packages/liferay-jest-junit-reporter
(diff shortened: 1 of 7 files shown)


jbalsas:
😍

PR #79, opened 2019-04-26
feat: add "wepback" subcommand to liferay-npm-scripts

wincent:
At the moment all this does is run "webpack", which will use the default "webpack.config.js" found in the current directory.

I have a liferay-portal PR under development which does some shuffling around of files in that repo in order to make this possible:

https://github.com/wincent/liferay-portal/pull/8

Next steps will include getting a usable "watch" command in place, and exploring whether we can offer a meaningful default config here. We'll also have the option of integrating the webpack build into the standard "liferay-npm-scripts build" command: the idea there will could be to run the webpack step if a config file is present; in order to make that happen though, will need to make some further changes to the upstream projects (for example, segments/segments-web is already using a custom invocation of `babel`, so would need to be adapted).

Based on https://github.com/liferay/liferay-npm-tools/pull/78, which will need to be merged before this one.

### DIFF
--- a/packages/liferay-npm-scripts/__fixtures__/scripts/webpack/sample/webpack.config.dev.js
+++ b/packages/liferay-npm-scripts/__fixtures__/scripts/webpack/sample/webpack.config.dev.js
@@ -1,0 +1,13 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+module.exports = {
+	entry: './index.js',
+	output: {
+		filename: 'bundle.js',
+		path: __dirname
+	}
+};
--- a/packages/liferay-npm-scripts/__tests__/scripts/webpack.js
+++ b/packages/liferay-npm-scripts/__tests__/scripts/webpack.js
@@ -1,0 +1,65 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const path = require('path');
+
+const FIXTURES = path.resolve(
+	__dirname,
+	'../../__fixtures__/scripts/webpack/sample'
+);
+
+describe('scripts/webpack.js', () => {
+	let cwd;
+	let spawnSync;
+	let webpack;
+
+	beforeEach(() => {
+		cwd = process.cwd();
+		jest.mock('../../src/utils/spawnSync');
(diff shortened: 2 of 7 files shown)


wincent:
Marking this as a draft to prevent accidental merging. It will need to be rebased and retargeted once #78 goes in.

jbalsas:
just merged #78

wincent:
Rebased and retargeted against "master".

wincent:
What do you think of this middle-ground, @jbalsas? `liferay-npm-scripts build` will automatically try to run webpack if a config file is present (no config), and we keep `liferay-npm-scripts webpack` so that we have a place to put our `liferay-npm-scripts webpack --watch` command. As no new configuration is needed, this is independent of what we decide to do with https://github.com/liferay/liferay-npm-tools/pull/76.

I updated https://github.com/wincent/liferay-portal/pull/8 to show how this would work in liferay-portal.

bryceosterhaus:
More of a larger question for this, but how come we are using webpack? Won't this mean that our dev environment will be isolated from the actual portal? I know for the segment editor, it was strange because all development was done in an isolated webpack-dev-server but then they would have to deploy to actually use it in the liferay context and there would sometimes be issues from the webpack bundler vs our portal bundler.

wincent:
@bryceosterhaus:

> but how come we are using webpack

I don't know the backstory on the frontend-js-web use cases, but I can comment on the other two:

- analystics-client-js is a standalone bundle for use outside of liferay-portal by third-parties: for a while it lived in a separate repo but was folded in. It will probably always be somewhat of an odd one out.
- segments-web needs webpack to work around a design constraint of our own bundler: segments-web needs a third-party library that contains a circular dependency and which our bundler is not designed to accommodate; our workaround there is to use webpack to create a bundle just for that third-party library, which then gets fed into the rest of the build process. segments-web also uses webpack to get a "watch" mode which is pretty slick and has apparently sped up their development workflow considerably.

frontend-js-web also has a watch mode: don't know if that's part of the original motivation for using it there. But I know we'd like to make "watch" functionality more broadly available and webpack could end up being a part of that.

@jbalsas knows the rationale a ~~bit~~lot more deeply though.

jbalsas:
LGTM, let's keep this moving and work on the presets story to to see where we land and what else might be missing!

wincent:
Ok, I'll merge this and cut a release. Then I'll update the remaining things on https://github.com/wincent/liferay-portal/pull/8 (hopefully won't need any further changes on this side, at least for now.... #FamousLastWords.)

PR #80, opened 2019-05-16
 feat: add class-properties and object-rest-spread support

wincent:
We are already using these in a number of places in the liferay-portal repo, and they are almost certain to be enshrined as-is in a future version of the standard:

- rest-spread is at stage 4: https://github.com/tc39/proposal-object-rest-spread
- class-properties are at stage 3 and implemented in Babel, Node, Chrome etc: https://github.com/tc39/proposal-class-fields


wincent:
Going to merge this so I can move forward, but will loop back to address any feedback that may come in later.

PR #81, opened 2019-05-16
feat: provide standard version of babel-loader

wincent:
So that modules in the liferay-portal repo can rely on it being present rather than pulling it in independently. I think liferay-npm-scripts is the right place for it because all webpack usage in the liferay-portal is happening via this package anyway.


wincent:
Trivial so going to merge it.

PR #82, opened 2019-05-16
feat: always clean up temporary Babel config files, and make webpack builds use them as well

wincent:


wincent:
In celebration of the National Day of Ninja-Merges, I've tested this in liferay-portal so I am going to ship it. I need to follow up with a fix to the `deepMerge` functionality, which seems busted (before this PR) in the module that I'm currently testing.

Issue #83, opened 2019-05-16
Update array merge algorithm to work well adequately with Babel


wincent:
In testing https://github.com/liferay/liferay-npm-tools/pull/82 I've discovered at least one case where the array-merge algorithm we're using does the wrong thing:

Given "user config" from segments-web:

```
{
	"presets": [
		"@babel/preset-react"
	],
	"plugins": [
		["module-resolver", {
			"root": ["./src/main/resources/META-INF/resources/js/"],
			"alias": {
				"test": "./test/js/"
			}
		}]
	]
}
```

and merging with the defaults from this project:

```
{
	"presets": ["@babel/preset-env"],
	"plugins": [
		"@babel/proposal-class-properties",
		"@babel/proposal-object-rest-spread"
	]
}
```

we get:

```
 { presets: [ '@babel/preset-env', '@babel/preset-react' ],
  plugins:
   [ [ 'module-resolver', [Object] ],
     '@babel/proposal-object-rest-spread' ] }
```

which, as you can see, is missing one of the plug-ins and breaks the build.

As a reduced test case,:

- merging `['a', 'b', ['c']]`
- and `['d', 'e']`
- yields: `['d', 'e', ['c'], 'b']`

The cause is our use of the pre-2.0 merge algorithm, removed [here](https://github.com/TehShrike/deepmerge/pull/77) and described in the changelog as "a complicated thing".

The post-2.0 algorithm is basically just a concat, which is close to what we'll need here. I'm going to make another merge algorithm specifically adequate for our Babel-merging use case here.

PR #84, opened 2019-05-17
feat: add Babel merge strategy to get-merged-config.js

wincent:
As described in the related issue, our merge implementation is using the pre-2.0 version of the array merge algorithm in the "deepmerge" package. They removed it because it was a "complicated thing".

In the case of merging Babel configs it verily breaks them, because you can specify plugins and presets as either `'name'` or `['name', options]`. Here's a reduced example of some surprising behavior:

- merging `['a', 'b', ['c']]`
- into `['d', 'e']`
- yields: `['d', 'e', ['c'], 'b']`

ie. `'a'` disappears into the void.

See the test cases included in this commit for a real-life example taken from segments-web. The old algorithm would cause the `@babel/plugin-proposal-class-properties` plugin to vanish.

The implementation here doesn't intend to handle all possible edge cases (for example, it doesn't try to match plugins that have been written in different forms such as "@babel/plugin-x" and "x"), but it covers the ones we are most likely to see in liferay-portal and it is well tested.

I chose to write a narrow implementation like this rather than pull in a third-party dependency like "babel-merge", because I like to keep our dependency footprint small and we actually kind of want to discourage complicated Babel overrides in liferay-portal in any case, so it's in our interests to have something that isn't too flexible.

Closes: https://github.com/liferay/liferay-npm-tools/issues/83

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/deep-merge.js
+++ b/packages/liferay-npm-scripts/src/utils/deep-merge.js
@@ -35,14 +35,125 @@ function combineMerge(target, source, options) {
  */
 const overwriteMerge = (destinationArray, sourceArray) => sourceArray;
 
+function getItemDescription(item) {
+	try {
+		return JSON.stringify(item);
+	} catch (error) {
+		// Could be a circular reference, but we're unlikely to ever get here
+		// because the deepmerge package itself will die first.
+		return `[unstringifiable item: ${error}]`;
+	}
+}
+
+function getBabelName(item) {
+	if (typeof item === 'string') {
+		return item;
+	} else if (Array.isArray(item) && typeof item[0] === 'string') {
+		return item[0];
+	} else {
+		throw new Error(
+			`getBabelName(): malformed item ${getItemDescription(item)}`
+		);
+	}
+}
+
+function getBabelOptions(item) {
+	if (typeof item === 'string') {
+		return null;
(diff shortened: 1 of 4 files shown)


wincent:
Going to merge this in so I can move forward with my liferay-portal pull. Will check back for follow-up items later. 👋

packages/liferay-npm-scripts/src/utils/deep-merge.js
```diff
@@ -35,14 +35,125 @@ function combineMerge(target, source, options) {
...
+			return merge.all(items, {
+				arrayMerge: combineMerge
+			});
+
+		case MODE.OVERWRITE_ARRAYS:
+			return merge.all(items, {
+				arrayMerge: overwriteMerge
+			});
+
+		case MODE.BABEL:
```
bryceosterhaus: Seems a little weird at first to make an exception for just babel here, could just be me though. However, the API is clear and makes sense with the whole `customMerge` option with `deepmerge`. Good idea with the `MODE` API.

wincent: If you can think of anything more general than this we should implement it. Might be challenging though to find something that knows how to merge arrays where elements like `'foo'`  and `['foo', {option: true}]` actually refer to the same "thing".

Issue #71, opened 2019-04-09
Recover original .npmbundlerrc file when bundler fails


izaera:
Right now, if the build fails, sometimes the modified `.npmbundlerrc` is left and the original one is lost. It would be great if the script could recover from such situation and restore the original `.npmbundlerrc`.

Alternatively, we could modify the `liferay-npm-bundler` (in addition to the scripts) so that it could receive the path to the `.npmbundlerrc` and that way we would execute it with a temporary one and leave the project's one untouched.

wincent:
> Alternatively, we could modify the `liferay-npm-bundler` (in addition to the scripts) so that it could receive the path to the `.npmbundlerrc` and that way we would execute it with a temporary one and leave the project's one untouched.

I think that would be great. I was pretty horrified — only exaggerating a *little* bit — when I saw how config merging worked in this repo by overwriting and then restoring the config files.

The "nice" thing about mutating the config is that it is left on the filesystem for you to inspect when things go wrong, but I think we can have that same benefit with a temporary config file too if we just emit the path to it in the console output.

PR #85, opened 2019-05-17
fix: recover original .npmbundlerrc after failures (#71)

wincent:
Unlike what I did with Babel in #82, just add a `try`/`finally` wrapper. I wanted something more generic and reusable for Babel because we use it in 3 places, but we only use liferay-npm-bundler in one place, so something simpler and more direct is called for.

Test plan: install in liferay-portal, add a `throw` after bundling, and run the `yarn build` in a module with an .npmbundlerrc file (eg. frontend-js-web). See the original config file is restored despite the error.

Closes: https://github.com/liferay/liferay-npm-tools/issues/71

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/build.js
+++ b/packages/liferay-npm-scripts/src/scripts/build.js
@@ -39,17 +39,19 @@ function compileBabel() {
  * `liferay-npm-bundler` executable.
  */
 function runBundler() {
-	moveToTemp(CWD, '.npmbundlerrc');
+	try {
+		moveToTemp(CWD, '.npmbundlerrc');
 
-	const RC_PATH = path.join(CWD, '.npmbundlerrc');
+		const RC_PATH = path.join(CWD, '.npmbundlerrc');
 
-	fs.writeFileSync(RC_PATH, JSON.stringify(BUNDLER_CONFIG));
+		fs.writeFileSync(RC_PATH, JSON.stringify(BUNDLER_CONFIG));
 
-	spawnSync('liferay-npm-bundler');
+		spawnSync('liferay-npm-bundler');
 
-	fs.unlinkSync(RC_PATH);
-
-	removeFromTemp(CWD, '.npmbundlerrc');
+		fs.unlinkSync(RC_PATH);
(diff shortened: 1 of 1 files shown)


packages/liferay-npm-scripts/src/scripts/build.js
```diff
@@ -39,17 +39,19 @@ function compileBabel() {
...
+		moveToTemp(CWD, '.npmbundlerrc');
 
-	const RC_PATH = path.join(CWD, '.npmbundlerrc');
+		const RC_PATH = path.join(CWD, '.npmbundlerrc');
 
-	fs.writeFileSync(RC_PATH, JSON.stringify(BUNDLER_CONFIG));
+		fs.writeFileSync(RC_PATH, JSON.stringify(BUNDLER_CONFIG));
 
-	spawnSync('liferay-npm-bundler');
+		spawnSync('liferay-npm-bundler');
```
bryceosterhaus: If the command fails at this point, how does this resolve? Won't we have two `.npmbundlerrc` files? One from line 47, and the original being restored in line 53.

wincent: Based on my testing, it puts everything back the way it was (with one caveat, noted below):

https://github.com/liferay/liferay-npm-tools/blob/d23683598ebde1132b2d88fc69c848dd5931ac4e/packages/liferay-npm-scripts/src/utils/move-to-temp.js#L15-L44

- Lines 25-26 of `move-to-temp.js` check for `$PWD/TEMP-.npmbundlerrc` and rename it back to `$PWD/.npmbundlerrc` (overwriting the config that we just wrote). If there is no such file (ie. if `moveToTemp()` didn't create one), nothing happens, which means that the config that was written out will stay there and show up in the `git status` output as an untracked file.
- Line 29 *would* check for config in `$PWD/package.json` and restore it, but does nothing here because we're not passing a `packageKey`.

So, a more complete solution would remove the created  `.npmbundlerrc` in the case that one didn't exist beforehand, but #71 was just asking that we not mangle an pre-existing file, so I didn't go that far.

Issue #89, opened 2019-05-23
Provide better error messages for duplicate plug-ins


wincent:
We recently added `@babel/plugin-proposal-class-properties` and `@babel/plugin-proposal-object-rest-spread` to our default config. If somebody creates a `.babelrc` with the same plugins but different names (possible because Babel allows for various shorthand notations), then we will end up passing both through and Babel will complain about the duplicates.

Example case of this and [the fix here](https://github.com/arboliveira/liferay-portal/pull/742/commits/a6ac0133391f8508eea117d1241912c58da75628) (ie. removing the duplicate). Sadly, Babel's error message isn't helpful:

```
Error: Duplicate plugin/preset detected.
If you'd like to use two separate instances of a plugin,
they need separate names, e.g.

  plugins: [
    ['some-plugin', {}],
    ['some-plugin', {}, 'some unique name'],
  ]
```

because it doesn't say which plug-ins are duplicated. And even if it did, it wouldn't be easy for people to troubleshoot because we dynamically create a `.babelrc` at runtime, so they'll probably inspect what they have on disk and wonder why there are being told there's a duplicate.

We have two options here to make a better experience:

- Normalize plug-in names to shorthand format before merging, thus silently eliminating duplicates.
- Normalize plug-in names to shorthand format, and if duplicates found, print an actually helpful error message instead.

I think the latter is better, because otherwise we may wind up with redundant config in our repo. It will be harmless, but still redundant. I'd rather people get a clear error that tells them how to remove the redundancy.

bryceosterhaus:
> Normalize plug-in names to shorthand format, and if duplicates found, print an actually helpful error message instead.

+1

PR #90, opened 2019-05-23
feat: enforce canonical preset and plugin names (#89)

wincent:
As described in the related issue, Babel has various optional shorthand conventions for identifying plugins and presets, which means that it is possible to refer to the same module under different names. Babel will then complain about duplicates with an error message which isn't very helpful:

    Error: Duplicate plugin/preset detected.
    If you'd like to use two separate instances of a plugin,
    they need separate names, e.g.

      plugins: [
        ['some-plugin', {}],
        ['some-plugin', {}, 'some unique name'],
      ]

It doesn't mention the actual names that are duplicated, or even if it's a preset or a plugin that it's complaining about. This, combined with the fact that we dynamically generate a ".babelrc" at runtime, means that this is very confusing for people not familiar with the tooling; they may look at their on-disk ".babelrc" and "package.json" wonder what duplicates it is talking about.

We could silently normalize the names and merge them, but I think it's better to do what we do in this commit: normalize to detect duplicates and issue an error if one is found; that way, we avoid redundant config from finding its way into the repo.

Closes: https://github.com/liferay/liferay-npm-tools/issues/89

### DIFF
--- a/packages/liferay-npm-scripts/__tests__/utils/deep-merge.js
+++ b/packages/liferay-npm-scripts/__tests__/utils/deep-merge.js
@@ -287,5 +287,64 @@ describe('deepMerge()', () => {
 				);
 			}).toThrow(/malformed item/);
 		});
+
+		it('complains about non-shorthand "@babel/plugin-foo"', () => {
+			expect(() => {
+				deepMerge(
+					[{plugins: []}, {plugins: ['@babel/plugin-foo']}],
+					deepMerge.MODE.BABEL
+				);
+			}).toThrow('expected "@babel/foo"');
+		});
+
+		it('complains about non-shorthand "@org/babel-plugin-foo"', () => {
+			expect(() => {
+				deepMerge(
+					[{plugins: []}, {plugins: ['@org/babel-plugin-foo']}],
+					deepMerge.MODE.BABEL
+				);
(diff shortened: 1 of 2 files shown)


wincent:
FWIW, this is the kind of thing you'll see in the console if you run a build in portal with a bad plugin or preset name:

```
$ yarn build
yarn run v1.15.2
$ liferay-npm-scripts build
/Users/greghurrell/code/portal/liferay-portal/modules/node_modules/liferay-npm-scripts/src/utils/deep-merge.js:83
                                        throw new Error(
                                        ^

Error: checkBabelName(): expected "@babel/preset-react", got "@babel/react"
    at Object.entries.reduce (/Users/greghurrell/code/portal/liferay-portal/modules/node_modules/liferay-npm-scripts/src/utils/deep-merge.js:83:12)
...
$ yarn build
yarn run v1.15.2
$ liferay-npm-scripts build
/Users/greghurrell/code/portal/liferay-portal/modules/node_modules/liferay-npm-scripts/src/utils/deep-merge.js:83
                                        throw new Error(
                                        ^

Error: checkBabelName(): expected "@babel/proposal-class-properties", got "@babel/plugin-proposal-class-properties"
...
 ```

Hopefully that is clear enough?

Issue #88, opened 2019-05-21
passing env params for testing


matuzalemsteles:
I do not know if by default the tests should have the env of `test` or we should just inherit the envs. This is required for Metal testing inside the portal.

wincent:
@matuzalemsteles Are you referring to `NODE_ENV` or something else?

matuzalemsteles:
Oh sorry for not being clear with this, just the `NODE_ENV`.

wincent:
Setting `NODE_ENV` to `test` seems pretty reasonable to me, and Jest [should do it by default](https://raw.githubusercontent.com/facebook/jest/4ba05b7d4c5a9f8acd75fec529da2e2543422a81/docs/GettingStarted.md):

> Jest will set `process.env.NODE_ENV` to `'test'` if it's not set to something else.

Seems we're setting it to "production" [here](https://github.com/liferay/liferay-npm-tools/blob/ab2cc17d473637f7bd034c7940c0d7b2bde1a428/packages/liferay-npm-scripts/src/scripts/test.js#L9). @bryceosterhaus: do you remember your rationale for doing that?

matuzalemsteles:
@wincent I think setting the `process.env` in this way will not influence the Jest call, I think we would have to configure the env of [spawnSync](https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options) I'm not sure about this, the documentation is not very clear.

wincent:
@matuzalemsteles I think the environment is inherited automatically; try this in the `node` REPL:

```
process.env.ZZZZZ = 'hi'
console.log(require('child_process').spawnSync('/usr/bin/env').stdout.toString())
```

You'll see:

```
[ a bunch of env variables]
ZZZZZ=hi
```

bryceosterhaus:
@wincent I had gotten the idea from react-scripts, [here](https://github.com/facebook/create-react-app/blob/e7a2d6168a3a93ea58bafdc851eb2bbda20d5ce6/packages/react-scripts/scripts/build.js#L12). I think we can follow the same pattern and isolate it to the `build` script only though. The main idea for this would be to ensure the proper NODE_ENV for babel.

matuzalemsteles:
Thanks @wincent seems to work fine. @bryceosterhaus for the test case, do we need it in `production`? it may not be necessary for testing.

bryceosterhaus:
[...]

PR #91, opened 2019-05-23
feat: set NODE_ENV only for builds (#88)

wincent:
This will allow Jest, for example, to use its normal default ("test").

Test plan: Use `yarn add -W path/to/liferay-npm-scripts` to install this version in the liferay-portal repo's "modules/" workspace root. Then change to a module that uses "liferay-npm-scripts" (eg. "segments/segments-web") and run `yarn build`; see:

```
$ liferay-npm-scripts build
Using NODE_ENV: production
Successfully compiled 39 files with Babel.
...
```

Then run `env NODE_ENV=development yarn build` and see:

```
$ liferay-npm-scripts build
Using pre-existing NODE_ENV: development
Successfully compiled 39 files with Babel.
...
```

Closes: https://github.com/liferay/liferay-npm-tools/issues/88

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/set-env.js
+++ b/packages/liferay-npm-scripts/src/utils/set-env.js
@@ -1,0 +1,38 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+function log(message) {
+	// eslint-disable-next-line no-console
+	console.log(message);
+}
+
+/**
+ * Sets NODE_ENV to the supplied `env`, unless the user has already explicitly
+ * provided a value.
+ *
+ * To override, run:
+ *
+ * ```
+ * env NODE_ENV=development liferay-npm-scripts subcommand
+ * ```
+ *
+ * Babel will consult NODE_ENV if BABEL_ENV is not defined, but you can
+ * override Babel specifically with:
+ *
+ * ```
+ * env BABEL_ENV=development liferay-npm-scripts subcommand
+ * ```
+ */
+function setEnv(env) {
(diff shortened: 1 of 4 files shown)


matuzalemsteles:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

packages/liferay-npm-scripts/src/utils/set-env.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+function log(message) {
```
bryceosterhaus: Whats the idea behind abstracting `console.log` here? Just to avoid the `no-console`?

Also, we should probably just remove the `no-console` rule.

wincent: Yep that's the sole idea. So I have exactly one `console.log` and the smallest suppression.

matuzalemsteles: Maybe it would not be better to disable the `no-console` only in that file?

bryceosterhaus: I'd probably vote to just remove the `no-console` rule altogether for this project. It also feels like a bit of overkill to abstract here. 

In the end it isn't a huge deal to me though, fine merging with it as is.

wincent: As a follow-up I made this: https://github.com/liferay/liferay-npm-tools/pull/92

With that there is only one `console.log` in the entire project, so I probably wouldn't touch the rule.

PR #92, opened 2019-05-24
refactor: teach getMergedConfig() and index to reject invalid types

wincent:
Based on discussion about use of `console.log` in:

https://github.com/liferay/liferay-npm-tools/pull/91

I did a quick audit and found that there are only two other places in this package where we use `console.log` and both should probably throw errors instead of just logging a message and proceeding.

For example, current behavior:

    $ liferay-npm-scripts formatz
    'formatz' is not a valid command for liferay-npm-scripts.
    $ liferay-npm-scripts
    'undefined' is not a valid command for liferay-npm-scripts.

(Both exit with status code 0.)

Behavior after this commit:

    $ liferay-npm-scripts formatz
    /path/to/index.js:46
                    throw new Error(
                    ^

    Error: liferay-npm-scripts requires a valid command (build, format, lint, test, theme, webpack)
    $ liferay-npm-scripts
    /path/to/index.js:46
                    throw new Error(
                    ^

    Error: liferay-npm-scripts requires a valid command (build, format, lint, test, theme, webpack)

(Both exit with status code 1.)

I changed a `switch` to an object so that I could auto-generate the list of commands, and also deleted some dead code on the way; we were creating a temporary dir but never using it.

In any case, now we have exactly one place in the code base that uses `console.log`.

### DIFF
--- a/packages/liferay-npm-scripts/src/index.js
+++ b/packages/liferay-npm-scripts/src/index.js
@@ -4,14 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const fs = require('fs');
 const minimist = require('minimist');
-const path = require('path');
-const rimraf = require('rimraf');
-
-const CWD = process.cwd();
-
-const TEMP_PATH = path.join(CWD, 'TEMP_LIFERAY_NPM_SCRIPTS');
 
 module.exports = function() {
 	const ARGS_ARRAY = process.argv.slice(2);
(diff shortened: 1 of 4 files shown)


packages/liferay-npm-scripts/src/index.js
```diff
@@ -4,14 +4,7 @@
...
  */
 
-const fs = require('fs');
 const minimist = require('minimist');
-const path = require('path');
-const rimraf = require('rimraf');
-
-const CWD = process.cwd();
-
-const TEMP_PATH = path.join(CWD, 'TEMP_LIFERAY_NPM_SCRIPTS');
```
matuzalemsteles: I realized this being created in conjunction with the merged files, maybe we should still keep and add the babel and jest files inside that folder. And I think we should add this folder to the `.gitignore` of the Portal.

wincent: How it currently works:

1. We make a backup (if needed) with names like `TEMP-.babelrc` by [`moveToTemp()`](https://github.com/liferay/liferay-npm-tools/blob/394aec503ef5d5178beab4769ff9379af25ab303/packages/liferay-npm-scripts/src/utils/move-to-temp.js#L57) in the current working directory.
2. We create a new file (eg. `.babelrc`) in the current working directory with the merged config (ie. standard filename, no configuration).
3. If there is an error, we put things back the way they were.

What you're proposing, if I understand it correctly:

1. Create ignored temp folder.
2. Put merged config in there.
3. Run tools with additional switches to make them look for that config in a non-standard location.
4. If there is an error, (clean up? or leave for inspection?)

I kind of hate the whole backup-overwrite-then-switch-back pattern, but I am not sure whether all of the tools we run from `liferay-npm-scripts` know how to accept custom config paths yet (for example, does `liferay-npm-bundler`?). Also there are some edge-casey things that might get unpleasant if we move config files away from their standard locations: for example, I can pass a custom config path to `webpack`, and it will look for a Babel config file in the usual places without any changes, but if we want it to respect Babel settings from a custom location we'll have to figure out how to force it to do so. As much as I dislike the file mutation that we're doing, it may be the simpler alternative.

bryceosterhaus: >  but I am not sure whether all of the tools we run from liferay-npm-scripts know how to accept custom config paths yet

This is the reason I went with the temp config files. I believe it was bundler, and csf that did not accept a custom config path.

matuzalemsteles: Hm, yeah I had forgotten that the tools behind must know this information.

The concern I have about modifying the original `.babelrc`, `jest...` files is that people will usually modify something there and we are always modifying those files and leaving everything inline this is pretty annoying sometimes.

PR #96, opened 2019-05-31
feat: replace check-source-formatting with prettier (#93)

### LINKED ISSUE #93: Swap out CSF for prettier
We got approval to move portal's js to be formatted with prettier rather than csf. See https://github.com/brianchandotcom/liferay-portal/pull/73576

To go this route we need to make the change in `liferay-npm-scripts` and then submit a pull to portal with the updated version and running format. The prettier config will be located here in `liferay-npm-scripts` and the rules for that can be seen here, https://github.com/brianchandotcom/liferay-portal/pull/73576/files#diff-7f33ec48d01d2b893151bb61f68b930a.

The `.prettierrc` would be added to root of liferay-portal. This is for text editors to automatically pick up the prettier config.

We should also slowly transition some of the rules we had in CSF to our eslint config.

bryceosterhaus:
Thought I would try this real quick and send a PR and see what you guys think.

A few notes:
- Removed `.soy` and `.jsp` files from formatting and linting since prettier doesnt handle those. Not sure what we want to do about that.
- Replaced our `getUserConfig` logic with [cosmiconfig]() since all of our configs fall under that pattern. Wasn't positive we wanted to do this, but I thought it would free us up from maintaining that part ourselves. I know you're allergic to dependencies though, @wincent.
- I added a `silent` arg for spawnProcess, also not sure about this. When prettier catches an improperly formatted file, it ends with an exit code of 1, making our output pretty sloppy. This also raises the question of if this is also necessary for CI failures, wasn't sure if we just keep the sloppy. Compare the two images below, one without the `silent` feature and one with.

With:
Screen Shot 2019-05-31 at 10 51 19 AM

Without:
Screen Shot 2019-05-31 at 10 52 43 AM

Anyways, I think thats all, let me know what you guys think.


bryceosterhaus:
disregard the open/close. Was hoping I could make this a "Draft" PR.

packages/liferay-npm-scripts/src/utils/spawnSync.js
```diff
@@ -17,8 +17,9 @@ function getDescription(command, args) {
  * @param {string} command Path to bin file
  * @param {Array=} args=[] List of string arguments
  * @param {Object=} options={} Options to pass to spawn.sync
+ * @param {boolean} silent=false Should swallow non-error exit codes.
  */
-module.exports = function(command, args = [], options = {}) {
+module.exports = function(command, args = [], options = {}, silent = false) {
```
wincent: If you want to add another option, it should go in the `options` object rather than being tagged on the end. You can "pluck" it out from the others with destructuring:

```javascript
const {silent, ...spawnOptions} = options;
```

Having said that, I think what we really want here is still to exit with a non-zero status (because that's the expected behavior in any of these tools) but we want the output to be less noisy, which we can achieve by printing the exception message only rather than the entire stack trace, which is the ugly bit.

packages/liferay-npm-scripts/src/scripts/lint.js
```diff
@@ -14,11 +16,18 @@ const LINT_PATHS = getMergedConfig('npmscripts').lint;
...
+	const args = [
+		'--config',
+		CONFIG_PATH,
+		fix ? '--write' : '--check',
+		...LINT_PATHS
+	];
+
+	spawnSync('prettier', args, {}, true);
+
+	fs.unlinkSync(CONFIG_PATH);
```
wincent: We'll want a `try`/`finally` here so that we don't leave a turd 🤭 on the the filesystem when/if things go wrong.

wincent:
Applied some commits on top. Most notably, linting produces non-zero exit code on finding a formatting issue (same as `csf` used to do), but we don't spew a stack trace anymore, so it looks like this:

liferay-portal____________node_modules__bin_liferay-npm-scripts

This is pretty close to your first screenshot, @bryceosterhaus, but with one extra line showing the error message.

bryceosterhaus:
Those changes sound good to me. Thanks for updating those, @wincent!

PR #99, opened 2019-06-03
doc: recommend using a recent version of NodeJS

wincent:
Because we don't use Babel to make these packages and they get run in the context of liferay-portal without transpilation, we should advise people to use a recent version so as not to choke on the various "modern JS" features that we use pretty freely in this codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,12 @@
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


wincent:
"no logic changes" 😂

PR #101, opened 2019-06-03
feat: provide a way to skip linting/formatting entirely

wincent:
In integrating `prettier` with liferay-portal I have found modules that have "checkFormat" scripts defined even though they don't have any source code (for example, wrapper modules like frontend-js-metal-web, which is basically just declares dependencies but doesn't contain any committed code of its own under "src/"):

    "scripts": {
            "build": "liferay-npm-scripts build",
            "checkFormat": "liferay-npm-scripts lint",
            "format": "liferay-npm-scripts format"
    },

For these, `prettier` will complain about not finding any files that match the default globs. You would think that you could define an `npmscripts.config.js` file, then, that removes the globs, but that won't work either (because `prettier` will complain that you didn't pass any globs).

I am going to explore if there is any downside to removing those "checkFormat" and "format" scripts, but I can also understand why we might want to be able to run `yarn format` in any module and not have it blow up, so that's what this commit is about: it makes it possible to configure an empty set of globs, in which case we won't invoke `prettier` at all.

Along the way, made a tweak to ensure that our behavior matches what we claim in the documentation (details in e954c27).

### DIFF
--- a/packages/liferay-npm-scripts/README.md
+++ b/packages/liferay-npm-scripts/README.md
@@ -36,7 +36,7 @@ Do you need to use `liferay-npm-bridge-generator`? Just add a `.npmbridgerc` fil
 liferay-npm-scripts lint
 ```
 
-Lint calls `check-source-formatting` for the globs specified in your `npmscripts.config.js` configuration. Or default preset seen [here](./src/presets/standard/index.js#L25-L32)..
+Lint calls `prettier` for the globs specified in your `npmscripts.config.js` configuration. Or default preset seen [here](./src/presets/standard/index.js#L25-L32)..
 
 ### format
 
@@ -44,7 +44,7 @@ Lint calls `check-source-formatting` for the globs specified in your `npmscripts
 liferay-npm-scripts format
 ```
 
-Format calls `check-source-formatting` with the `--inline-edit` flag for the globs specified in your `npmscripts.config.js` configuration. Or default preset seen [here](./src/presets/standard/index.js#L17-L24).
+Format calls `prettier` with the `--write` flag for the globs specified in your `npmscripts.config.js` configuration. Or default preset seen [here](./src/presets/standard/index.js#L17-L24).
 
 ### test
 
--- a/packages/liferay-npm-scripts/__tests__/scripts/lint.js
+++ b/packages/liferay-npm-scripts/__tests__/scripts/lint.js
@@ -1,0 +1,70 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+describe('scripts/lint.js', () => {
+	let cwd;
+	let lint;
+	let spawnSync;
+	let temp;
+
+	beforeEach(() => {
+		cwd = process.cwd();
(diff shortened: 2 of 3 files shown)


wincent:
Just to provide some more context on this, [JSONPackagJSONCheck.java](https://github.com/liferay/liferay-portal/blob/73398a9014750949e90ea4c383f1d25c2faf2f00/modules/util/source-formatter/src/main/java/com/liferay/source/formatter/checks/JSONPackageJSONCheck.java) in [liferay-portal](https://github.com/liferay/liferay-portal) kind of forces our hand here: my understanding is that it is going to run on any project that has a package.json file and require that the "format" and "checkFormat" keys are defined and that they exactly match "liferay-npm-scripts format" and "liferay-npm-scripts lint" exactly. This is why even places where it makes no sense to have these keys (like [in modules/package.json](https://github.com/liferay/liferay-portal/blob/73398a9014750949e90ea4c383f1d25c2faf2f00/modules/package.json#L24-L27)), we have them.

There are exactly three places in liferay-portal where I'll have to add an otherwise useless "npmscripts.config.js" file, which isn't too bad given that the initial diff adding prettier will be almost 90,000 lines.

Downsides? If we ever add source to those locations it will only be linted/formatted if we remember to update or remove the overrides. That seems fine: we'll notice soon enough, and the alternatives aren't any better (eg. if we could remove the "liferay-npm-scripts format" lines, we'd also have to remember to add those back).

bryceosterhaus:
> Downsides? If we ever add source to those locations it will only be linted/formatted if we remember to update or remove the overrides. That seems fine: we'll notice soon enough, and the alternatives aren't any better (eg. if we could remove the "liferay-npm-scripts format" lines, we'd also have to remember to add those back).

Just one thought on this. IMO, it seems better to remove the `checkFormat` lines, otherwise I think a dev will be more likely to see the `checkFormat` script and assume "oh its formatting, so my code must be fine." Whereas, if we remove the `checkFormat`, it'll be clear that there its not actually formatting since the script doesn't exist.

julien:
@bryceosterhaus I also agree that throwing an error, a warning or anything similar would be a better option, we discussed this with @wincent at the office and that was our initil idea too; but after seeing [JSONPackageJSONCheck.java](https://github.com/liferay/liferay-portal/blob/73398a9014750949e90ea4c383f1d25c2faf2f00/modules/util/source-formatter/src/main/java/com/liferay/source/formatter/checks/JSONPackageJSONCheck.java), we thought it might be better to make the builds pass rather than fail, and as you can see we have various checks that are hard to bypass.

@wincent will correct me if I'm mistaken, but the three packages in which we had to add empty globs are "our" packages so we'll just have to be carefull if we ever add source code in there.

bryceosterhaus:
I am all for making the builds pass, and maybe I am mistaken, but does removing the `checkFormat` script cause a failed build?

wincent:
In the end, we didn't have to create three packages with empty globs: it ended up being two, and the third one has non-empty globs: it's [modules/npmscripts.config.js](https://github.com/liferay/liferay-portal/blob/4f9138b8649e50dd0048fed3a6495d5851c511e1/modules/npmscripts.config.js#L4-L10), which handles formatting for all the files not otherwise covered by a project-local config. (I'm about to change that file though anyway, so don't pay too much attention to what's inside it at that rev.)

Because the top-level `checkFormat` ends up matching pretty much all source files, I think we don't have to worry about devs having a false sense of security — it will be a actual security... 😂

Issue #95, opened 2019-05-31
Add "test/**/*.js" configuration to standard preset


wincent:
Internal ticket for this: https://issues.liferay.com/browse/LPS-95744

We don't have many frontend tests in liferay-portal yet, but as we add them, we're finding that we need to configure explicitly otherwise they don't get linted/formatted.

So, task here is to:

- Add standard "test/**/*.js" paths to the default preset.
- Cut a liferay-npm-scripts release.
- Make sure all tests in liferay-portal are in standard location.
- Update to new version to get the presets, then delete the config.

See the JIRA ticket for more info.

wincent:
FWIW: "test" isn't used super commonly in liferay-portal, but it is used in some places. For example, in the "modules/apps" directory, I ran `find . -type d -depth 2 -exec sh -c 'echo {}; ls {} | egrep "src|test"' \;` to get a list of all apps and show whether they had "src" and/or "test" directories. Full output is [here](https://gist.github.com/wincent/ca73209de915545f7aba729c21165527). You can see that:

- Out of 1092 projects (`find . -type d -depth 2`)
- 1065 have "src" directories (`find . -type d -name "src" -depth 3`)
- Only 10 have "test" directories (`find . -type d -name "test" -depth 3`)

Note that that is just looking for the directory names; it doesn't try to scope this only to frontend-projects, so some of those may not have any JS or SCSS at all. To get a sense of where our tests are I also ran `find . -path '*/test/*' -name '*.js'` — that got 181 files ([full output](https://gist.github.com/wincent/b64c3b3d5336c8b72cf06b0e70a0e984)). Most of them are under top-level "test", some under "src/test" (which would be caught by the "src" globs, and a few scattered elsewhere (like false positives under "build/"). [Our Jest config](https://github.com/liferay/liferay-npm-tools/blob/1f30289692be55cbd3a508ea09fdd8de14221971/packages/liferay-npm-scripts/src/config/jest.json) uses "test" in it's glob, so it will find all of those anyway (it excludes "build", "classes" etc).

wincent:
Confirmed with @jbalsas and @julien that top-level "tests/" is the place we want things to be.

- PR updating the globs: https://github.com/liferay/liferay-npm-tools/pull/103
- PR removing some dead tests that were in the wrong place: https://github.com/brunobasto/liferay-portal/pull/1036
[...]

PR #103, opened 2019-06-04
feat: expand lint/format globs to cover "test" (#95)

wincent:
As described in the related issue, we determined that:

- We want test code in liferay-portal to reside in a top-level "test" directory inside each module.
- Currently there are only two modules which don't follow that pattern:
  - One is dead code which I am proposing to remove here: https://github.com/brunobasto/liferay-portal/pull/1036
  - The other can be moved; currently testing a PR that does that here: https://github.com/wincent/liferay-portal/pull/28

Closes: https://github.com/liferay/liferay-npm-tools/issues/95

### DIFF
--- a/packages/liferay-npm-scripts/src/presets/standard/index.js
+++ b/packages/liferay-npm-scripts/src/presets/standard/index.js
@@ -14,6 +14,10 @@ module.exports = {
 		input: 'src/main/resources/META-INF/resources',
 		output: 'classes/META-INF/resources'
 	},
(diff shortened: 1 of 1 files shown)


wincent:
Tested this in liferay-portal and it works. Final diff is going to look (something like) this: https://gist.github.com/wincent/089d0e535e34dab592ff7666be7852e7 (warning: 30k-line gist which will crush your browser; [raw version](https://gist.githubusercontent.com/wincent/089d0e535e34dab592ff7666be7852e7/raw/8832a20291dd726fea79ecaabfb3c84418a570f5/patch.diff) is probably better).

Issue #109, opened 2019-06-07
Make it possible to run ESLint


wincent:
For overall goal, see: https://issues.liferay.com/browse/LPS-96565

This specific issue, though, corresponds to the first step: https://issues.liferay.com/browse/LPS-96566

We want a way to separately run ESLint inside portal so that we can get a sense of how many lint violations it is going to catch. Depending on the volume/nature of the violations, that will guide our next steps.

PR #112, opened 2019-06-07
feat: add subcommands for running ESLint (#109)

wincent:
Not finished yet, but pushing this because I have to leave in a moment and will pick it up later. Still to do:

- [x] Actually provide default config.
- [x] Test in liferay-portal.
- [ ] ~Refactor to eliminate duplication.~ (going to wait and see a bit on this one until I've tested more things in liferay-portal).

Closes: https://github.com/liferay/liferay-npm-tools/issues/109


wincent:
This is now at the point where I can actually test it in liferay-portal, so updating this as an "early look". Probably will find some issues though so please don't merge it yet.

wincent:
I've tested this in liferay-portal now and it looks like everything that used to work still works, plus we have the ability to start running and addressing ESLint issues now, so I am going to merge this and will loop back if anybody has any suggestions.

Issue #94, opened 2019-05-29
Add Storybook


thektan:
Related to [this thread](https://liferayproductdev.slack.com/archives/C03CAD9EF/p1558474909005200) on slack.

> Entry devDependencies is not allowed.

We've been working on a new module (not merged in master yet) and we've been using Storybook (https://storybook.js.org/) for the frontend dev environment which has been great for working/testing components in isolation.

Here's a link to the current storybook setup we have: https://github.com/thektan/liferay-portal/tree/LPS-89524_Search_Tuning_Result_Rankings/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/stories

- It includes loading in language keys (currently only the current module's language keys are translated) https://github.com/thektan/liferay-portal/blob/LPS-89524_Search_Tuning_Result_Rankings/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/stories/webpack.config.js#L27-L30
- It also [proxies portal](https://github.com/thektan/liferay-portal/blob/LPS-89524_Search_Tuning_Result_Rankings/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/stories/middleware.js#L13) to retrieve resources like [icons](https://github.com/thektan/liferay-portal/blob/LPS-89524_Search_Tuning_Result_Rankings/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/stories/index.js#L32) and [styles](https://github.com/thektan/liferay-portal/blob/LPS-89524_Search_Tuning_Result_Rankings/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/stories/preview-head.html).

Since `devDependencies` are not allowed, would Storybook be something we'd want to add? Or possibly some other alternative that doesn't require the entire portal to need a dependency that possibly only a single module will use.

wincent:
Some "meta" stuff, not specifically about Storybook:

---

Just to reflect some of the context from that Slack thread here, seeing as not everyone may have permission to view that link, I'll paraphrase @jbalsas who was explaining the "no devDependencies" stance:

- We have a lint in place to stop new devDependencies being added because we want a centralized/standard decision making procedure for the build process and what tools get brought in. Motivation here is:
  - Performance: as devDependencies get added download/build times increase.
[...]

PR #113, opened 2019-06-08
feat: add "storybook" command (#94)

thektan:
Here's an early first pass of adding a `liferay-npm-scripts storybook` command. I'm sharing this now in case anyone notices any issues/improvements.

It currently works by copying the storybook config files in `src/storybook` into a `TEMP-storybook` directory on the current working directory and removing it when storybook is closed.

## Current concerns I have
- ~Portal modules would still need to add `devDependencies` for `@storybook/*` packages since methods are being used to write stories.~ https://github.com/liferay/liferay-npm-tools/pull/113#issuecomment-500514418 https://github.com/liferay/liferay-npm-tools/pull/113/commits/d6ba93b2da596641f5a9ceff159e74083bb5834d
- ~Resources (css/icons) currently depend on an instance of portal running (maybe provide some type of warning or fallbacks if one isn't found?)~ https://github.com/liferay/liferay-npm-tools/pull/113#issuecomment-500514418

## Todo
- [x] Explore http-server to serve resources
- [x] Make portal url configurable (localhost:8080)
- [x] Test with other modules (segments-web)
- [x] Use storybook presets to bundle webpack and addons (https://storybook.js.org/docs/presets/introduction/)
- [x] Use node temp dir for storing buillt storybook configs https://github.com/liferay/liferay-npm-tools/pull/113#issuecomment-502586873


jbalsas:
Hey @thektan, thanks for sending this over! Looks great as a first iteration!

Regarding your concerns:

> Portal modules would still need to add devDependencies for @storybook/* packages since methods are being used to write stories.

This won't be allowed. The goal of providing this here is to ensure we develop consistent documentation for our UI components and that all teams do so without harming other developers productivity. We should pick the plugins that make sense for all of us and stick to those. Door is always open to suggest more (or less) depending on how things evolve, but always upon consensus.

> Resources (css/icons) currently depend on an instance of portal running (maybe provide some type of warning or fallbacks if one isn't found?)

It should be fine to iterate on this and define proper workflows. In fact, it could be that we start a simple http-server to serve the static resources or something like that so we don't need a real running portal.

Please, @thektan, keep working on the missing items and signal us when ready. @bryceosterhaus is on PTO this week, but @wincent might find some time to go over this.

packages/liferay-npm-scripts/src/scripts/storybook.js
```diff
@@ -0,0 +1,46 @@
...
+
+	const args = [
+		'--port',
+		STORYBOOK_CONFIG.port,
+		'--config-dir',
+		TEMP_DIR_NAME
+	];
+
+	// Run the cleanup function when storybook is exited with `ctr+c`.
+	process.on('SIGINT', cleanup);
```
wincent: Should also probably wrap the `spawnSync` call in `try`/`finally` so that we do clean-up if an exception is thrown too.

packages/liferay-npm-scripts/src/storybook/addons.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+ /**
+ * Order determines the order they will be displayed in the tabs.
+ */
```
wincent: Not sure if this is the case, but whitespace looks broken here.

packages/liferay-npm-scripts/package.json
```diff
@@ -19,14 +19,22 @@
...
+		"@storybook/addon-knobs": "^5.1.3",
+		"@storybook/addon-viewport": "^5.1.3",
+		"@storybook/react": "^5.1.3",
 		"babel-loader": "8.0.6",
 		"cosmiconfig": "^5.2.1",
 		"cross-env": "^5.2.0",
 		"cross-spawn": "^6.0.5",
+		"css-loader": "^2.1.1",
 		"deepmerge": "^3.0.0",
+		"fs-extra": "^8.0.1",
```
wincent: Are we sure "fs-extra" is justifying its weight here? It's not obvious to me why we need to create a temporary directory for the stories in the first place — can you help me understand how that need arises?

thektan: When starting up storybook you need to specify **a directory** which contains the storybook config files like `config.js` and `addons.js`. It currently doesn't seem like there's a way to point to a specific file or configure what the names of specific files are.

fs-extra was added to manipulate directories, but I suppose another possibility is adding custom functions like in the ones in `utils`. I found this gist that does directory manipulation: https://gist.github.com/tkihira/3014700

wincent: So why can't we just pass `path.join(__dirname, '../storybook')`? It's still not clear to me why we would need to make a copy.

wincent: (And just to be explicit, you'd obviously need to change the `require('../stories');` thing in the config.js file to load from the current working directory instead of a path relative to the config.js.)

thektan: Ooo I see, you're right! I was wondering if there was a way to run the command pointing to the storybook directory in `liferay-npm-scripts`. Didn't realize I had it there already.

Since storybook is the one calling `config.js`, would you have any ideas about how I'd load from the current working directory? Possibly somehow import the `cwd` path into it? Or modify the file in the `storybook` script?

thektan:
@jbalsas 
> > Portal modules would still need to add devDependencies for @storybook/* packages since methods are being used to write stories.
> 
> This won't be allowed. The goal of providing this here is to ensure we develop consistent documentation for our UI components and that all teams do so without harming other developers productivity. We should pick the plugins that make sense for all of us and stick to those. Door is always open to suggest more (or less) depending on how things evolve, but always upon consensus.

I suppose we could export the specific storybook addons used? In the portal modules using storybook they would be used like:
```javascript
import {
    storybookReact, 
    storybookAddonKnobs, 
    storybookAddonA11y
} from 'liferay-npm-scripts';
```
---
> > Resources (css/icons) currently depend on an instance of portal running (maybe provide some type of warning or fallbacks if one isn't found?)
> 
> It should be fine to iterate on this and define proper workflows. In fact, it could be that we start a simple http-server to serve the static resources or something like that so we don't need a real running portal.

The main motivation was to make the dev environment as close to portal styles as possible since while I was working outside of portal context, the component would end up looking a bit off after placing it inside portal because of portal global styles being applied on top of base `clay-css`. 

`http-server` is a good idea. It'd have to compile css files too, but it seems doable. I'll explore that option as well.

packages/liferay-npm-scripts/package.json
thektan: Oh so changing it to `require(cwd + '/stories');` does work.

wincent: That shouldn't be happening: other scripts in liferay-npm-scripts rely on `process.cwd()` returning the right directory (ie. the one containing `modules/apps/*/package/json`). Minor nit: we should use `path.join()` instead of `+`.

thektan: The storybook `config.js` seems to be called within the browser (assuming since `process.title = 'browser'`), so I've set a new environment variable `process.env.STORYBOOK_CWD` within storybook's webpack to use within `config.js`.

(Also updated to use `path.join()`)

thektan:
# Update

- Since storybook required a `.babelrc` in the storybook config directory, I created an ignored  `/build/storybook` directory in `liferay-npm-scripts` to keep the compiled `.babelrc`. https://github.com/liferay/liferay-npm-tools/pull/113/commits/436909730c8c9e70cc86b52729b866cc4031ffd4
- I also merged multiple `Language.properties` files together and placed the generated file inside of `build/storybook` as well. https://github.com/liferay/liferay-npm-tools/pull/113/commits/4978415c75519bc1fb7ed91f1321a0057c5b2d57
  - I'm not sure if this `/build` directory should be placed within `liferay-npm-scripts` or in the `cwd` instead though. Or if the generated files should just be placed directly in `liferay-npm-scripts/src/storybook` and git ignored.
- Instead of an `http-server`, storybook supports loading static files from a directory, so I've configured it to point to the root portal directory where it can reference specific files. (https://github.com/liferay/liferay-npm-tools/commit/0e528d437cc78183945668e91ba865ce95013790#diff-1cb1d2dc5cf78ec0e057315b44bc24b3R24) I still think proxying portal is a good idea to keep in case components reference a `/o/*` resource. If we had a `http-server` it seems like we'll have to maintain it to keep it in line with how portal resources are served and at that point it seems like we can directly reference portal instead. https://github.com/liferay/liferay-npm-tools/pull/113/commits/0e528d437cc78183945668e91ba865ce95013790

thektan:
@wincent @jbalsas I added documentation about how using the storybook script would work. When you get the chance, let me know how it sounds (mostly from a dev attempting to use it point-of-view). 🙂

https://github.com/liferay/liferay-npm-tools/blob/9b7d0bbcb5da8f7b120ba504aa45a2c05c446e20/packages/liferay-npm-scripts/src/storybook/README.md

wincent:
> I'm not sure if this /build directory should be placed within `liferay-npm-scripts` or in the `cwd` instead though. Or if the generated files should just be placed directly in `liferay-npm-scripts/src/storybook` and git ignored.

Using an internal directory of something that gets installed into `node_modules` seems a bit naughty IMO. I think there are two places that you could make an argument for, for storing ephimeral config:

- In the `cwd`: you could argue that you want to leave a copy around for inspection in case things go wrong (and conversely, you might consider leaving stuff around to be a bug, and make sure you clean it up using `finally` blocks — this is actually the approach that we've mostly taken in this repo historically).
- Using the OS temp-dir facilities: eg. `fs.mkdtempSync(path.join(os.tmpdir(), 'storybook-')) — this gives you something well isolated that doesn't need to be Git ignored, and which will get automatically cleaned up by the OS at some later point. You can still have introspectability in the event of a failure by logging the path where the config files are being created, which is the best of both worlds because you can debug things but you don't have to worry about cluttering up the `git status` output.

Personally, I like the second option because it feels like using the right tool for the job.

thektan:
@wincent I'll make changes to use the temp-dir facilities since it seems more appropriate (as the functions are named). Thanks for making me aware of that option.

I'll also add a log to where the directory ends up being created at.

One current concern I have about that though is if you started and stopped storybook a bunch of times, would you possibly end up with hundreds of new files? But I suppose this could be solved by removing the tmp dir after stopping storybook.

wincent:
> One current concern I have about that though is if you started and stopped storybook a bunch of times, would you possibly end up with hundreds of new files?

@thektan: Yes, if you start and stop, you'll wind up with a new directory each time. However, the semantics of `os.tmpdir()` are that the operating system is expected to clean out its contents for you. You can write whatever you want in there and know that it will get garbage collected automatically in the future. In practice, that usually means "on next reboot" but it is platform dependent. It should be fine unless you are writing gigabytes of stuff in there, which we aren't doing.

As an example, on my (macOS) machine, `os.tmpdir()` currently returns `/var/folders/yc/1x2y2qld1g95tvtjpd89hdrm0000gp/T`. For security, each time I boot the OS will give me a different, unpredictably-named folder. My machine hasn't been rebooted for 26 days, so I have 648 MB of stuff in that directory, spread across 4,738 subfolders. Here are the last few:

```
yarn--1560847408066-0.8514897881684278
yarn--1560847408414-0.908374001680412
yarn--1560847413609-0.8732727139399932
yarn--1560847417165-0.30375650731740556
yarn--1560847448466-0.42758731439612063
yarn--1560847466918-0.28203989351078573
yarn--1560847970689-0.2002349073342815
yarn--1560847970689-0.42899571370498735
yarn--1560847970689-0.7561983026163224
yarn--1560847970689-0.7934091678358626
```

In short, I wouldn't worry about cleaning up the temporary directories.

packages/liferay-npm-scripts/package.json
```diff
@@ -37,6 +44,9 @@
 		"minimist": "^1.2.0",
 		"prettier": "1.18.2",
 		"rimraf": "^2.6.2",
+		"sass-loader": "^7.1.0",
+		"sort-keys": "^3.0.0",
```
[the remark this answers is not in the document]
thektan: Good catch! I must have accidentally added it after a rebase.

thektan:
@wincent I realized I was doing something else hacky where `middleware.js` and `webpack.config.js` were referencing files outside of its build directory. So when moving to the temp directory the `require` statements wouldn't resolve. Would there be a way of still referencing the npm modules inside `liferay-npm-scripts`? Create a webpack for the webpack? 😅

wincent:
> Would there be a way of still referencing the npm modules

Maybe. What files are you trying to `require` exactly? One way to communicate with the storybook config files would be to send an environment variable that tells it where to resolve node_modules from.

For example, in the call to `spawnSync`:

```javascript
spawnSync('start-storybook', args, {
    env: {
        LIFERAY_NPM_SCRIPTS_NODE_MODULES: path.resolve(path.join(__dirname, '..', '..')),
    },
});
```

Then if the webpack.config.js is copied to some arbitrary location, it could do:

```javascript
const NODE_MODULES = process.env.LIFERAY_NPM_SCRIPT_NODE_MODULES || process.cwd();

const thing = require.resolve('thing', {paths: [NODE_MODULES]});
```

Or something to that effect (untested, typed directly into browser). Without knowing more about what your changes are looking like, not sure exactly whether that's the best solution. There might be a simpler one. Can you share a work-in-progress?

thektan:
@wincent All the changes in this PR are up-to-date with my work in progress.

thektan:
@wincent Pushed up the change that uses the temp directory.

When starting up this is a snippet of the error:
```
ERR! Error: Cannot find module 'http-proxy-middleware'
ERR!     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
ERR!     at Function.Module._load (internal/modules/cjs/loader.js:562:25)
ERR!     at Module.require (internal/modules/cjs/loader.js:690:17)
ERR!     at require (internal/modules/cjs/helpers.js:25:18)
ERR!     at Object.<anonymous> (/private/var/folders/7s/x15jvhps42q73dkpcc5sjc8r0000gn/T/storybook-BDSiw7/middleware.js:7:15)
ERR!     at Module._compile (internal/modules/cjs/loader.js:776:30)
ERR!     at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
ERR!     at Module.load (internal/modules/cjs/loader.js:653:32)
ERR!     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
ERR!     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
ERR!     at Module.require (internal/modules/cjs/loader.js:690:17)
ERR!     at require (internal/modules/cjs/helpers.js:25:18)
...
```

wincent:
@thektan: So it looks like this is what is happening:

1. You have a file with both relative and non-relative imports:

    ```javascript
    const proxy = require('http-proxy-middleware');
    
    const getMergedConfig = require('../../src/utils/getMergedConfig');
    ```

2. You copy that file to some arbitrary location (eg. `/tmp/blah`).
3. Node can't resolve `http-proxy-middleware` because there is no local "node_modules" directory, nor is there one in any of the ancestors of `/tmp/blah`.
4. Even if it could resolve it, it will fail to resolve `../../src/utils/getMergedConfig`, because there is no file at that location either.

I see about 12 non-relative `import`s/`require`s in the "storybook" directory, and 1 relative one. I believe you can take care of the first 12 by [setting `NODE_PATH`](https://nodejs.org/api/modules.html) in the `spawnSync` environment so that Node will search the local `node_modules` directory and find those modules (see "Loading from the global folders" in the linked document).

The relative one, I think you can get rid of it by doing the call to `getMergedConfig` from inside `liferay-npm-scripts`, before you call `spawnSync`: just write the resulting config into the same temporary directory.

I know this seems complicated, but I think it's worth seeing if this can be made to work without dumping files into a subdirectory of `node_modules`.

thektan:
Thanks for the helpful suggestions @wincent! I made the changes reflecting your comments. I think this is finally ready for a full review.

packages/liferay-npm-scripts/src/scripts/storybook.js
```diff
@@ -0,0 +1,148 @@
...
+	);
+
+	const LANG_PATHS = [PORTAL_LANG_PATH, ...paths];
+
+	// Used for keeping track of the current file being processed to show the
+	// correct file that caused an error.
+	let currentLangPath;
+
+	try {
+		const bufferArray = LANG_PATHS.filter(path => fs.existsSync(path)).map(
```
wincent: Probably shouldn't shadow the outer `path` binding here. I'd make this `langPath` or similar, or better yet, skip the variable entirely and do `LANG_PATHS.filter(fs.existsSync)`. You'd still need to do `.map(langPath => { ... }` though.

packages/liferay-npm-scripts/src/scripts/storybook.js
```diff
@@ -0,0 +1,148 @@
...
+			path.join(buildPath, 'Language.properties'),
+			buffer.toString('utf8')
+		);
+	} catch (e) {
+		log(`Failed to read lang key file: ${currentLangPath}`);
+
+		// Write an empty file to prevent liferay-lang-key-dev-loader from
+		// breaking the build if no file is found.
+		fs.writeFileSync(path.join(buildPath, 'Language.properties'), '');
+	}
```
wincent: One idea here, you could avoid the `let currentLangPath` stuff by moving the `try`/`catch` inside the loop. Apart from wanting to avoid an unnecessary temporary variable, keeping the contents of a `try` as narrowly scoped as possible seems like a good idea as a general practice.

```javascript
let buffer = Buffer.alloc(0);

LANG_PATHS.filter(fs.existsSync).forEach(langPath => {
  try {
    buffer = Buffer.concat([
      buffer,
      fs.readFileSync(langPath),
      Buffer.from('\n')
    ]); 
  } catch (error)
    log(`Failed to read lang key file "${langPath}" due to error: ${error}`);
  }
});

fs.writeFileSync(
  path.join(buildPath, 'Language.properties'),
  buffer.toString('utf8')
);
```

As far as I can tell, that would mean you'd get (at least) your empty file no matter what, so you wouldn't even need the comment calling it out.

I also wonder though whether all this needs to be done using buffers explicitly. Would just strings suffice? eg.

```javascript
let output = [];

LANG_PATHS.filter(fs.existsSync).forEach(langPath => {
  try {
    output.push(fs.readFileSync(langPath).toString());
  } catch (error)
    log(`Failed to read lang key file "${langPath}" due to error: ${error}`);
  }
});

fs.writeFileSync(
  path.join(buildPath, 'Language.properties'),
  output.join('\n')
);
```

packages/liferay-npm-scripts/src/scripts/storybook.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/**
+ * Starts a storybook server for testing frontend components.
+ */
+module.exports = function() {
```
wincent: One practice that I like for improving stack traces is to always use named functions (except for cases where arrow functions are obviously better). ie. you could make this `module.exports = function storybook() { ...` or:

```javascript
function storybook() {
  // ...
}

module.exports = storybook;
```

packages/liferay-npm-scripts/src/scripts/storybook.js
```diff
@@ -0,0 +1,148 @@
...
+const getMergedConfig = require('../utils/getMergedConfig');
+const log = require('../utils/log');
+const spawnSync = require('../utils/spawnSync');
+
+const BABEL_CONFIG = getMergedConfig('babel');
+const STORYBOOK_CONFIG = getMergedConfig('npmscripts').storybook;
+
+const NODE_PATHS = [
+	path.join(__dirname, '../../../../node_modules'),
+	path.join(__dirname, '../../node_modules')
```
wincent: 😱 — let's keep this one a little secret just between us. 😂

packages/liferay-npm-scripts/src/scripts/storybook.js
thektan: Tested out with strings instead and it works. 👌 Looks much cleaner so I'll go with that approach. I'll keep in mind the structuring of the try/catch blocks though!

packages/liferay-npm-scripts/src/scripts/storybook.js
thektan: If I make a function like this...does it make it better or worse? 😅
```
function getNodeModulesPath(packageName) {
	const pathArray = __dirname.split('/');

	return pathArray
		.slice(0, pathArray.indexOf(packageName) + 1)
		.join('/')
		.concat('/node_modules');
}
```

And then using it:
```
const NODE_PATHS = [
	getNodeModulesPath('liferay-npm-scripts'),
	getNodeModulesPath('liferay-npm-tools')
];
```

thektan:
@wincent Oops, there was a rebase conflict so I just removed the yarn.lock. I'll regen the changes for it. 👍

packages/liferay-npm-scripts/src/scripts/storybook.js
wincent: I like the honesty of the original version better. Fewer things to go wrong, too.

thektan: 👍 I felt somewhat similar after comparing them too. I'll keep it as is.

bryceosterhaus:
@thektan fyi with yarn, if you have a merge conflict on the `yarn.lock`, you can just run `yarn` again and it automatically fixes conflicts.

thektan:
After adding in `import {fetch} from 'frontend-js-web'` to a portlet's components, this causes storybook to error out because it can't resolve `frontend-js-web`. I'll need to work on a way to somehow resolve modules imported from other portlet modules.

wincent:
@thektan: I would expect that the `NODE_PATH` hack should be enough to make `import from 'frontend-js-web'` work, because our use of Yarn workspaces means that frontend-js-web exists as a symlink in the top-level `node_modules` folder. I wonder if it is using some kind of custom resolution... Do you have a stack trace of when the error is happening? If it is happening in webpack, we might have to set up some custom [resolve](https://webpack.js.org/configuration/resolve/) config to tell it where to look for modules.

thektan:
@wincent Sadly even with adding the NODE_PATH or configuring it in webpack, the module still can't be resolved. I'm assuming since the package is symlinked in `node_modules` it's not finding the module.

The error happening in webpack:
```
Module not found: Error: Can't resolve 'frontend-js-web' in '/Users/ktan/Liferay/master/portal/modules/apps/portal-search/portal-search-ranking-web/src/main/resources/META-INF/resources/js/utils'
 @ ./src/main/resources/META-INF/resources/js/utils/api.js 1:0-40 20:9-14
 @ ./src/main/resources/META-INF/resources/js/components/ResultsRankingForm.js
```

thektan:
Adding this to `webpack.config.js` will work, but `frontend-js-web` isn't transpiled so it still errors out. 

```js
	config.resolve.alias = {
		...config.resolve.alias,
		'frontend-js-web': path.join(
			PORTAL_ROOT,
			'./modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/index.es.js'
		)
	};
```

Error:
```
ERROR in /Users/ktan/Liferay/master/portal/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/compat/modal/Modal.soy 1:11
Module parse failed: Unexpected token (1:11)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> {namespace liferay.frontend.Modal}
|
...
```

I'm not sure if we'd want to attempt to build all the necessary dependencies (loading soy in this case) or mock `fetch` for now.

thektan:
Pushed up a change that mocks the frontend-js-web via webpack. 2a13eeb

wincent:
> but `frontend-js-web` isn't transpiled so it still errors out

😢

I don't know whether there will be a general solution for this (depending on other modules in liferay-portal in general), but solving this at least for frontend-js-web seems pretty essential because we end up depending on it pretty much everywhere.

That suggests to me either we come up with a comprehensive "fake" (like the one you started in that last commit), or we figure out how to transpile frontend-js-web... Just spitballing here, but we do at least know that we can build it (ie. with `liferay-npm-scripts build`) and we know where it gets built to (ie. `./modules/apps/frontend-js/frontend-js-web/classes/META-INF/resources/index.es.js`), so we could potentially map it there.

@julien: any ideas?

jbalsas:
I still don't understand why the module would not be found in our current workspace scenario... maybe we should try to isolate this first before going down the rabbit hole...

In any case, I think @wincent suggestion of running the build process and then resolving to the built folder makes sense. Maybe we could wrap this logic in a custom webpack loader to simplify the resolution/building process?

thektan:
@wincent @jbalsas As a heads up, I'll be on PTO from July 10-23 so I can continue working on this when I get back. But anyone is also welcome to add changes to this if they wish. Thanks for everyone's help so far though!

If we build `frontend-js-web` we'd still have to deal with the calls to `Liferay.Loader`, so we'd probably need to instead have some type of custom webpack like @jbalsas noted. But I don't know how maintainable that'll be or if soy components would be functioning properly. We could maybe start off by transpiling specific, commonly used functions like `fetch` and maybe mock the rest.

thektan:
@wincent @jbalsas I've decided to [mock the `frontend-js-web` module](https://github.com/liferay/liferay-npm-tools/blob/5059d126e1bde313a8cb4aa52e9b477f73c7a186/packages/liferay-npm-scripts/src/storybook/frontend-js-web.mock.js) in the meantime since I think this will be good for a first iteration. I'd like to revisit building the `frontend-js-web` in a separate ticket since I'd like to remove the storybook dev dependencies and configs from our WIP portlet before it is sent to upstream. It would also take (me) significant effort in figuring out a custom build process for the frontend-js-web and I don't want this storybook script forever in limbo when there is already benefit to having a place for a basic component sandbox.

Also as an additional note to `frontend-js-web` not resolving: It turns out that the issue isn't symlinked modules, but rather something within the module itself. I tested `import 'liferay-amd-loader'` which isn't symlinked and it also throws a `Module not found: Error: Can't resolve 'liferay-amd-loader'`. Doing `import 'liferay-amd-loader/build/loader/loader'` will not throw an error though.

wincent:
Thanks for your work on this @thektan. Hope your PTO was good. I just got back from PTO myself a day ago, so I'm in review mode and will see if I can get to this and get it merged Real Soon Now™; unless something catches fire I should be able to spend time on it this week.

wincent:
@thektan: thanks for your patience on this one. I think this one is ready to merge. I tested it in liferay-portal with the following tweaks pushed on top to your remote/branch:

- Instead of `<root>/stories/index.js`, use `<root>/test/stories/index.es.js` (because it's not currently viable to be in top-level like that).
- Update the examples to match.

I think we'll soon be making a `Liferay` mock for use in tests, so at that point we can review whether to switch your mock over to that, keep it as-is, or explore the alternative approach that we discussed earlier (of building and configuring webpack to use the build files).

PR #116, opened 2019-06-12
chore: remove sort-keys dependency

wincent:
Just to reduce our dependency footprint a little and remove some "clutter" from the source. We added this in 08fa78d03ebd197 but as far as I can tell there is no need for the keys to be sorted (the merged configs are never committed to the repo or inspected by a human, and we don't do anything order-sensitive with them either).

Note that we still have another copy of sort-keys in the yarn.lock, but that one is a transitive dependency so I'm going to ignore it for now.

### DIFF
--- a/packages/liferay-npm-scripts/package.json
+++ b/packages/liferay-npm-scripts/package.json
@@ -35,7 +35,6 @@
 		"minimist": "^1.2.0",
 		"prettier": "^1.17.1",
 		"rimraf": "^2.6.2",
-		"sort-keys": "^3.0.0",
 		"webpack": "^4.30.0",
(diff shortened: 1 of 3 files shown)


wincent:
@bryceosterhaus: if my analysis is off here and this actually is needed please feel free to revert when you get back from PTO.

bryceosterhaus:
TBH, I don't really remember why I initially added `sort-keys`... I remember it initially being necessary, but I it looks like it may not be necessary anymore. It _may_ have been for the old `eject` script, but not totally positive. I'm sure its fine to remove it, but we should keep an eye out for any issues that might arise from this.

PR #121, opened 2019-06-19
feat(scripts): make it possible to target "*.es.js" alone

wincent:
Two things here:

Firstly, we need a hack in order to be able to target just "*.es.js" files with ESLint without torpedoing the use of Prettier on all files.

Secondly, by design, apparently, ESLint won't let you have more than one .eslintignore file. It will only look in the current directory, unless you pass an explicit `--ignore-path` option. If you do that, or it finds an .eslintignore in the current directory, it will skip over any local "eslintIgnore" property that might exist in the package.json too.

However, we kind of need this ability because we run `liferay-npm-scripts lint` in two modes:

1. From the "modules/" root, where we lint everything in the repo.
2. From individual subproject directories, like "modules/apps/segments/segments-web".

In both cases, we want to provide some very broad config that applies to the whole repo -- eg. ignore all `*.js` files but lint all `*.es.js` ones; for more context on that see: https://issues.liferay.com/browse/LPS-97079 and related tickets.

But then we want to provide more granular overrides at the per-project level so that we can slowly tighten the screws and increase our lint coverage of the entire repo.

This is an inglorious hack, but hopefully the comments and code explain it sufficiently.

Tested this with [this diff applied in my liferay-portal checkout](https://gist.github.com/wincent/2513be083a8c273417d5d7747aaf9cdc).

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/lint.js
+++ b/packages/liferay-npm-scripts/src/scripts/lint.js
@@ -5,6 +5,7 @@
  */
 
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 
 const filterGlobs = require('../utils/filterGlobs');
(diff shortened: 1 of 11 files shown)


packages/liferay-npm-scripts/src/scripts/lint.js
```diff
@@ -22,6 +23,159 @@ const DEFAULT_OPTIONS = {
...
+		} else {
+			log(`Did not find expected .eslintignore at ${root}`);
+		}
+	}
+
+	const contents = ignores.join('\n');
+
+	const directory = fs.mkdtempSync(
+		path.join(os.tmpdir(), 'liferay-npm-scripts-')
+	);
```
wincent: Note: explicitly not cleaning up this file because the OS will reclaim anything we leave in `os.tmpdir()` at a later time, which might be convenient for inspection, and we don't have to worry about it cluttering up the `git status`.

bryceosterhaus: I never knew about `os.tmpdir()` before. Very cool.

wincent:
That's a good point. I think we're basically going to have to aim to work the ignore list down as close to empty as possible, because there is going to be some unfortunate gap depending on where you set your working directory.

We might consider adding a build step that synchronizes ignores in both directions, but I would wait and see what happens first.

Issue #105, opened 2019-06-04
Support async/await in jest tests


thektan:
When using async/await in tests the following error is thrown: 
```
ReferenceError: regeneratorRuntime is not defined
```

### Option 1
According to [Jest docs](https://jestjs.io/docs/en/tutorial-async#async-await):
> To enable async/await in your project, install @babel/preset-env and enable the feature in your babel.config.js file.

In `.babelrc`
```json
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "targets": {
          "node": "current"
        }
      }
    ]
  ]
}
```
(Sources: https://github.com/facebook/jest/issues/3126#issuecomment-483320742, https://github.com/babel/babel/issues/7601#issuecomment-374566429)

### Option 2
Installing `@babel/plugin-transform-runtime` and adding the following configuration to `.babelrc`
```json
{
  "env": {
    "test": {
      "plugins": ["@babel/plugin-transform-runtime"]
    }
  }
}
```
(Sources: https://github.com/facebook/jest/issues/3126#issuecomment-465926747, https://github.com/babel/babel/issues/8829#issuecomment-456524916)

### Option 3
Importing `babel-polyfill` to the jest `setup.js` file. Which would mean adding `babel-polyfill` as a new devDependency.
```javascript
import 'babel-polyfill';
```
(Source: https://github.com/babel/babel/issues/5085#issuecomment-277544677)

## Final Comments
I'm currently using Option 2 since it scopes the plugin to just the tests (where the issue is only occurring), but I'd like to hear everyone else's opinion since I haven't noticed any differences with any of these fixes besides Option 3 throwing a new error `Warning: Can't perform a React state update on an unmounted component` in the tests (but that's probably something I'd need to fix in the component/tests itself).

wincent:
Given that we currently use Node v10.15.1 in liferay-portal (which has native `async`/`await` support), it would be nice to see how far we can get with "Option 1", because it means Babel will skip some of the heavier-weight transforms that it does (including relying on the regenerator runtime).

thektan:
@wincent It looks like we could do something like this example from the [jest docs](https://jestjs.io/docs/en/getting-started):

> Jest will set process.env.NODE_ENV to 'test' if it's not set to something else. You can use that in your configuration to conditionally setup only the compilation needed for Jest, e.g.

```javascript
// babel.config.js
module.exports = api => {
  const isTest = api.env('test');
[...]

PR #138, opened 2019-06-27
feat: support async/await in jest tests (#105)

thektan:
I created a `babel.test.json` to separate the babel configurations between the test and build scripts. I didn't use the [`const isTest = api.env('test');` in `babel.config.js` example from the jest docs](https://jestjs.io/docs/en/getting-started#using-babel) so we could continue to use the `babelMerge` function in `deepMerge.js` as is.

But I'm also not sure what the implications would be if we only applied the following to the current `babel.json` instead of creating a separate `babel.test.json`.

```
{
	"presets": [
		[
			"@babel/preset-env",
			{
				"targets": {
					"node": "current"
				}
			}
		]
	],
...
```

*Also as an additional note, currently the only place that will have async/await in tests is still in a dev branch yet to be merged in master. https://github.com/thektan/liferay-portal/blob/LPS-96356-fix-frontend-tests/modules/apps/portal-search/portal-search-ranking-web/test/js/components/ResultsRankingForm.js

### DIFF
--- a/packages/liferay-npm-scripts/src/config/babel.json
+++ b/packages/liferay-npm-scripts/src/config/babel.json
@@ -3,5 +3,20 @@
 	"plugins": [
 		"@babel/proposal-class-properties",
 		"@babel/proposal-object-rest-spread"
+	],
+	"overrides": [
+		{
+			"test": "**/test/**/*.js",
+			"presets": [
+				[
+					"@babel/preset-env",
+					{
+						"targets": {
+							"node": "current"
+						}
+					}
+				]
+			]
+		}
 	]
 }
--- a/packages/liferay-npm-scripts/src/config/babel.json
+++ b/packages/liferay-npm-scripts/src/config/babel.json
@@ -12,7 +12,7 @@
 					"@babel/preset-env",
 					{
 						"targets": {
-							"node": "current"
+							"node": "10.15"
 						}
 					}
 				]


wincent:
Thanks @thektan. Would it be possible to do this without creating `babel.test.json` using [the "overrides" feature](https://new.babeljs.io/docs/en/next/babelrc.html#overrides)? Something like:

```diff
diff --git a/packages/liferay-npm-scripts/src/config/babel.json b/packages/liferay-npm-scripts/src/config/babel.json
index 94c14b0..64eb93a 100644
--- a/packages/liferay-npm-scripts/src/config/babel.json
+++ b/packages/liferay-npm-scripts/src/config/babel.json
@@ -3,5 +3,16 @@
 	"plugins": [
 		"@babel/proposal-class-properties",
 		"@babel/proposal-object-rest-spread"
+	],
+	"overrides": [
+		{
+			"test": "**/test/*.js",
+			"presets": [
+				"env",
+				{
+					"targets": {"node": "current"}
+				}
+			]
+		}
 	]
 }
```

(Untested — the "test" glob might not be exactly right — but that's the basic idea.)

thektan:
@wincent Neat! I used the same glob used in the [jest configs](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/config/jest.json#L7).

wincent:
Nice! Glad it ended up being a minimal change. Let's merge this and we'll see if it breaks all the JS tests in liferay-portal. 😂

wincent:
@thektan: I rebased your branch on top of current "master" and added one tweak on top (b12b6fc) just to make sure we don't end up depending on untranspiled features that don't exist in liferay-portal's version of Node (v10.15.1) yet. It has `async`/`await`, for example, but not public class fields (silly example, I know, as we're unlikely to use those in a test, but you get the idea).

PR #142, opened 2019-07-03
feat: handling globbing internally for ESLint

wincent:
Analog to what we did with Prettier in:

https://github.com/liferay/liferay-npm-tools/pull/141

I removed all the `findProjectIgnoreFiles()` stuff because we never ended up actually using it in liferay-portal, so it is unneeded complexity. (In the end we went with inline suppressions and created tickets to get rid of those.)

As before, tested in liferay-portal with a diff like this:

https://gist.github.com/wincent/dd173f939ba5b1c3b5906bffc4ba99a5

(see that one includes some formatting changes caused by our increased Prettier coverage).


wincent:
Tested this in liferay-portal, so will merge it and follow up on any feedback that comes in later.

PR #154, opened 2019-07-09
feat: teach findRoot() to work in master-private branch

wincent:
On "master" our root is "modules".

On "master-private" it is "modules/private".


wincent:
Obviously this is not an exhaustively rigorous check (for example, you could create a "yarn.lock" in a path like "foo/bar/baz/private/"), but it should be sufficient for all practical purposes.

PR #157, opened 2019-07-10
chore(scripts): tweak "master-private" concept of "root"

wincent:
This is the "Plan B" related to this comment:

https://github.com/brianchandotcom/liferay-portal-ee/pull/25312#issuecomment-509784609

As I explained there, our approach was to treat the two directories that contain "yarn.lock" files as separate little worlds (which, from Yarn's point of view, they are).

However, there is more than one way to skin a cat, and we can slightly tweak our definition of what "root" means for the purposes of linting and formatting (which are the only tasks where we are currently using `findRoot()`).

This commit makes the top (public) "modules" be considered the root and that's where we look up our ignore files. The files that I had added in that PR above are basically subsets of what is in the top-level files, simply because there is less stuff in "master-private" than master. The patterns in the top-level files should work fine.

Test plan: install this in "master-private" and recreate the PR above.


wincent:
cc @jbalsas

PR #160, opened 2019-07-11
feat(scripts): enable tests to have cross-project imports

wincent:
The "segments/segments-web" tests are currently broken because we moved `debounce()` into "frontend-js-web". This leads to two problems:

- Evaluating the `index.es.js` entry point of "frontend-js-web" causes us to transitively invoke a Soy template ("Modal.soy") which, in the context of "segments-web", has not been built, and we have no mapping to find.
- We can't even find `index.es.js` because we don't have a mapping that enables us to translate from the "main" property defined in the "package.json" of "frontend-js-web" to where the file really resides on disk (several levels buried down).

So, we add a transform that turns these ".soy" imports from other projects into harmless no-ops while still allowing projects to exercise their own actually-compiled-to-".soy.js" files, and we automatically set up mappings based on an analysis of the "workspaces" defined in the top-level "package.json".

I've been testing this in liferay-portal and it seems to handle all the edge cases I've found so far. I'll comment again once I've verified that every single module that uses `liferay-npm-scripts test` in "master" and "master-private" still work. A nice bonus will be that we can remove some custom mapping and transform config along the way.


wincent:
@andresfulla: This is the hackery I told you about that we need to fix segments/segments-web

@brunobasto: As I mentioned in one of the commits, I am going to open a ticket to create real "main" entry points in the DDM modules. I wasn't going to create them myself because I don't think I'm the right person to judge what is "public" and should be in the entry point and what is private.

@jbalsas: This is basically the approach that I told you about yesterday.

wincent:
Lints are going to fail on this because of dependency on #159.

wincent:
Just rebased on top of #159.

PR #165, opened 2019-07-15
feat(scripts): add ability to lint/format changed files only

wincent:
We think that the way we'll be getting passed the changed file list from the Java build tools in liferay-portal is going to be via an environment variable, so that's what is implemented here. Specifically, a comma-separated list passed via LIFERAY_NPM_SCRIPTS_CHANGED_FILES.

We chose this because we can read it conveniently from inside liferay-npm-scripts without having to pass the information down through multiple levels. Additionally, the length limits seem to be more liberal for environment variables than command line argument lists.

- On Windows, maximum environment variable size is either 31,767 or 8,191 characters, but on Linux is is effectively unlimited:

  https://stackoverflow.com/questions/1078031/what-is-the-maximum-size-of-an-environment-variable-value

- On Windows, the maximum command line string is supposedly 8,191 characters:

  https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string

  On other platforms, it is configurable but likely much higher:

  https://stackoverflow.com/questions/19354870/bash-command-line-and-input-limit

Using this command to get a crude estimate of path lengths in liferay-portal:

    find modules/apps -name '*.es.js' \
      -not -path '*/node_modules/*' \
      -not -path '*/classes/*' \
      -not -path '*/build/*' | \
    awk '{
      len=length($0);
      total+=len
    } END {
      printf("average: %d\n", total/NR);
    }'

that shows that the average path length for ".es.js" files is 120 characters, which means that if a limit of 8,191 applied then we could only expect to handle about 68 paths in the environment payload (on Windows), but wouldn't have to worry about it on Linux/macOS.

If this approach ends up not working out, we can either pass the upstream branch name instead and just compute the changed file list internally using Git commands, or we could pass the file list in via standard-input.

No matter what approach we end up using, we should be able to easily adapt the `filterChangedFiles()` function to work.

When running outside of the Java tooling, `yarn format` and `yarn checkFormat` will continue to work as before, operating without any filtering in place.


wincent:
Tested in liferay-portal with this: https://gist.github.com/wincent/784e1fa344115facbec942712ea6a1b2

Works on a subset; output of `ant format-source-current-branch` (note the file counts):

```
[exec] > Task :yarnFormatModules
[exec] yarn run v1.13.0
[exec] $ liferay-npm-scripts fix --production false
[exec] Prettier checked 21 files, found 0 files with problems
[exec]
[exec] ESLint checked 20 files, found 0 errors, found 0 warnings
[exec] Done in 6.27s.
[exec]
[exec] BUILD SUCCESSFUL in 10s
[exec] 1 actionable task: 1 executed
```

packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
```diff
@@ -0,0 +1,64 @@
...
+		'--diff-filter=ACMR',
+		'--name-only',
+		mergeBase,
+		'HEAD'
+	)
+		.split(/\0/)
+		.map(file => {
+			return file ? path.join(topLevel, file) : file;
+		})
+		.filter(Boolean);
```
wincent: In case you are wondering what this is for, it's because Git will append a trailing `NUL` byte after every name, so we'll get `foo\0bar\0baz\0`, so when we split we'll get `['foo', 'bar', 'baz', '']` and have to trim off the last one.

wincent:
Just confirmed that `ant format-source-all` correctly looks at everything:

```
[exec] $ liferay-npm-scripts fix --production false
[exec] Prettier checked 1450 files, found 0 files with problems
[exec]
[exec] /Users/greghurrell/code/portal/liferay-portal/modules/apps/hello-soy/hello-soy-navigation-web/src/main/resources/META-INF/resources/Navigation.es.js
[exec]   18:8  warning  'Footer' is defined but never used  no-unused-vars
[exec]   19:8  warning  'Header' is defined but never used  no-unused-vars
[exec]
[exec] /Users/greghurrell/code/portal/liferay-portal/modules/apps/hello-soy/hello-soy-web/src/main/resources/META-INF/resources/View.es.js
[exec]   18:8  warning  'Footer' is defined but never used  no-unused-vars
[exec]   19:8  warning  'Header' is defined but never used  no-unused-vars
[exec]
[exec] ✖ 4 problems (0 errors, 4 warnings)
[exec]
[exec] ESLint checked 551 files, found 0 errors, found 4 warnings
[exec] Done in 24.10s.
```

wincent:
Chatted with @jbalsas offline about this and we're going to give it a shot.

wincent:
https://github.com/brianchandotcom/liferay-portal/pull/75708 is the corresponding upstream pull.

bryceosterhaus:
Will this break if people are using this outside of liferay-portal?

wincent:
I would expect it to produce no behavioral changes outside of liferay-portal, because `LIFERAY_NPM_SCRIPTS_WORKING_BRANCH_NAME` will never be set in the environment. But, I don't know why anybody would run `liferay-npm-scripts` outside of liferay-portal; do you know about some use case like that @bryceosterhaus?

bryceosterhaus:
I think anyone that uses `gradle formatSource` will end up using this right? So I think the analytics team will essentially be using it.

Unless I am wrong about `formatSource` using these scripts

wincent:
I see what you mean. Yeah, in practice there are a few portal-ish places where we use or will use `liferay-npm-scripts`. The most obvious ones are liferay-portal ("master", "7.2.x") and liferay-portal-ee ("master-private"). We have some efforts underway to backport Yarn (and then `liferay-npm-scripts`) to other branches (eg. "7.1.x") and other repos (eg. commerce).

I think the base goal is for everything to work the same everywhere, and if it doesn't we'll fix it. In this specific instance, I think that the mechanisms defined in liferay-portal "master" (ie. the `portal-impl/build.xml` and `modules/util.gradle`) that this depends on end up getting applied to all the other places (via the overlay mechanism and/or by specific backports and cherry picks), and `liferay-npm-scripts` probably comes after that. And until it does, it will "degrade" gracefully if `LIFERAY_NPM_SCRIPTS_WORKING_BRANCH_NAME` is not set (it would just check/fix a bit more than it should).

bryceosterhaus:
Sounds good to me and thanks for the explanation. I was just looking at something in analytics recently and noticed they were using formatSource, so it was fresh in my head.

PR #164, opened 2019-07-15
feat(scripts): add @testing-library dependencies

wincent:
As a starting point, add `@testing-libary/react` and `@testing-library/jest-dom`, which are the two we'd discussed previously here:

https://github.com/liferay/liferay-npm-tools/issues/108

Test plan:

`yarn add -W --dev ~code/liferay-npm-tools/packages/liferay-npm-scripts` in liferay-portal checkoud, remove old deps (jest-dom and react-testing-library) from "modules/apps/segments/segments-web", update imports to use new libary names, run `yarn test`.

PR #171, opened 2019-07-16
feat(scripts): add @testing-library/user-event

wincent:
This should probably have been included in the initial PR that bundled our test dependencies (#164). This one was requested [here](https://github.com/liferay/liferay-frontend-guidelines/issues/49#issuecomment-510459340) and it seems like a worthy candidate for inclusion for a few reasons:

- It is an official package in the `@testing-library` organization, so we can expect a comparable level of support to that which the other `@testing-library` packages receive, and we've already opted to depend on (ie. trust) those.
- The package is a (literally) [very thin wrapper](https://github.com/testing-library/user-event/blob/251f287995c8a71/src/index.js) around the `fireEvent()` API, easily audited.
- The library encourages us to test what really matters (product behavior, often initiated by user activity) -- eg. typing into an input field -- with an expressive and ergonomic API, as opposed to having to deal with low-level browser implementation details, such as what order keydown/keypress/keyup/change events fire in.

One thing to note: until this PR lands upstream, this change adds a bit of duplication to the yarn.lock:

    https://github.com/testing-library/user-event/pull/136

Once that goes in we can update to remove the duplicate version of `@testing-library/dom`.

### DIFF
--- a/packages/liferay-npm-scripts/package.json
+++ b/packages/liferay-npm-scripts/package.json
@@ -22,6 +22,7 @@
 		"@babel/preset-react": "7.0.0",
 		"@testing-library/jest-dom": "4.0.0",
 		"@testing-library/react": "8.0.5",
+		"@testing-library/user-event": "4.1.0",
 		"babel-eslint": "^10.0.2",
 		"babel-loader": "8.0.6",
 		"cosmiconfig": "^5.2.1",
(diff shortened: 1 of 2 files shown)


wincent:
cc @andresfulla

wincent:
Green: https://travis-ci.org/liferay/liferay-npm-tools/jobs/559443711

PR #191, opened 2019-08-08
fix: adds missing clay and frontend-js-web dependencies configuration

jbalsas:
We currently have some missing configuration imports in our `preset-dev`. This causes some unnecessary modules to be processed.

Based on a quick estimate, it adds around 8s per module, so fixing this would provide a significant speed boost to build time.

**Test Plan:**
- Build a module in portal using `liferay-npm-scripts` that depends on `frontend-js-web`
- Assert that the resulting `classes/node_modules` does not contain the `frontend-js-web` dependencies (`liferay-amd-loader`, `svg4everybody`...)

### DIFF
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -92,7 +92,15 @@
 				"senna": ">=2.6.1"
 			},
 			"frontend-js-web": {
-				"/": ">=1.0.0"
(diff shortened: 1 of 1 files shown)


jbalsas:
closing in favour of https://github.com/liferay/liferay-npm-tools/pull/189 by @julien who should be taking care of his 🐱 instead of sending PRs ;)

wincent:
I copied your description over to the other PR. Thanks!

PR #192, opened 2019-08-08
fix: make generateSoyDependencies() work with lone dependencies

wincent:
If you had a single dependency like "metal-slider" then we were making a glob like `{path/to/directory/metal-slider}/src/**/*.soy`, which was failing to match anything at all.

For a "{a,b,c}"-style glob to work, it has to have at least two comma-separated items in it. "{a}" will not match "a".

So, in the case that the dependencies list only contains one item, use the item as-is; otherwise, wrap in "{"/"}" and join with commas.


wincent:
This fixes the issue that @jbalsas was seeing on his 7.0.x branch changes.

PR #195, opened 2019-08-12
chore!: update eslint-config-liferay to v6.0.0

jbalsas:


packages/liferay-npm-scripts/src/config/jest.config.js
```diff
@@ -7,17 +7,17 @@
...
 	modulePathIgnorePatterns: ['/__fixtures__/', '/build/', '/classes/'],
 	testMatch: ['**/test/**/*.js'],
 	testPathIgnorePatterns: ['/node_modules/', '<rootDir>/test/stories/'],
 	testResultsProcessor: 'liferay-jest-junit-reporter',
 	testURL: 'http://localhost',
 	transform: {
-		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js'),
-		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js')
+		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js'),
+		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js')
```
wincent: Hm. I am not sure about this one. I think it might use the "first match wins" approach, which means that before this change, `transformSoy.js` would have applied to ".soy" files, and `transformBabel.js` to everything else. After this change, it looks like the opposite.

jbalsas: You're probably right... we can err on the side of cauting here and just ignore sorting here and in `NORMALIZERS` as well

jbalsas: Pushed an additional commit restoring the plugin order and disabling eslint for those lines

packages/liferay-npm-scripts/src/config/jest.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
 	testResultsProcessor: 'liferay-jest-junit-reporter',
 	testURL: 'http://localhost',
 	transform: {
-		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js'),
-		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js')
+		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js'),
+		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js') // eslint-disable-line sort-keys
```
wincent: FWIW this is probably less obvious and more fragile than:

```javascript
		// eslint-disable sort-keys
		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js'),			'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js'),
		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js')
		// eslint-enable sort-keys
```

(in the event that the list grows over time). Same below, although I don't think it is going to grow much, if ever.

jbalsas: got it! wasn't aware of that syntax... I can push force with it, just to stay on the safe side

jbalsas:
Green it is!

PR #198, opened 2019-08-16
feat: force the use of a "test" NODE_ENV when running jest

wincent:
Due to recent changes in liferay-portal, we started setting an explicit NODE_ENV of "production" by default, which meant that Jest's usual behavior of assuming "test" no longer worked, and tests broke.

We resolved that in https://github.com/liferay/liferay-portal/commit/0741aabb5d201578a0 but in this commit we want to make ourselves 100% resilient to any future changes in the way `liferay-npm-scripts test` is invoked by always setting a NODE_ENV of "test", no matter what.

Here's what you see with no NODE_ENV:

    $ yarn test
    $ liferay-npm-scripts test
    Using NODE_ENV: "test"

And if NODE_ENV is "test":

    $ env NODE_ENV=test yarn test
    $ liferay-npm-scripts test
    Using NODE_ENV: "test"

And if NODE_ENV is anything else:

    $ env NODE_ENV=development yarn test
    $ liferay-npm-scripts test
    Overriding pre-existing NODE_ENV: "development" → "test"

Test plan: `yarn add` this package in portal and run tests with various NODE_ENV settings; see them pass.

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/test.js
+++ b/packages/liferay-npm-scripts/src/scripts/test.js
@@ -34,18 +34,16 @@ module.exports = function(arrArgs = []) {
 			NODE_ENV: 'test'
 		};
 
-		if (process.env.NODE_ENV !== 'test') {
-			log('Using NODE_ENV: "test"');
+		const {NODE_ENV} = process.env;
 
-			if (process.env.NODE_ENV) {
-				log(
-					'',
-					'Jest requires a NODE_ENV of "test", so the pre-existing',
-					`NODE_ENV of ${JSON.stringify(
-						process.env.NODE_ENV
-					)} was overridden`
-				);
-			}
+		if (!NODE_ENV || NODE_ENV === 'test') {
+			log('Using NODE_ENV: "test"');
+		} else {
+			log(
+				`Overriding pre-existing NODE_ENV: ${JSON.stringify(
+					NODE_ENV
+				)} → "test"`
+			);
 		}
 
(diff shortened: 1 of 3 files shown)


packages/liferay-npm-scripts/src/scripts/test.js
```diff
@@ -28,8 +29,29 @@ module.exports = function(arrArgs = []) {
 			buildSoy();
 		}
 
+		const env = {
+			...process.env,
+			NODE_ENV: 'test'
+		};
+
+		if (process.env.NODE_ENV !== 'test') {
+			log('Using NODE_ENV: "test"');
```
jbalsas: Should we log this anyways? I think that's what we do in the `build` script?

wincent: In the build script we log either:

    Using pre-existing NODE_ENV: "foo"

or:

    Using NODE_ENV: "production"

Reason I didn't log here if already "test" is just to keep the "noise" down, but I guess that's pointless and it won't hurt to log all the time. (If we do what I suggested earlier and stop passing "test"/"production" explicitly unless the user sets a concrete `nodejs.node.env`, then we'll end up logging all the time anyway.)

wincent:
Windows build stalled [here](https://travis-ci.org/liferay/liferay-npm-tools/jobs/572656432) but the other Windows build is [green](https://travis-ci.org/liferay/liferay-npm-tools/jobs/572656427), so this is probably fine to merge.

PR #202, opened 2019-08-19
feat(preset): add "/" import config for frontend-js-react-web

wincent:
In service of [LPS-99399](https://issues.liferay.com/browse/LPS-99399) we want to be offer a standard way to mount React apps in liferay-portal which will take care of common needs like:

- Establishing context(s).
- Subscribing to "~destropyPortlet~destroyPortlet" events and automatically calling `unmountComponentAtNode()`.
- etc

The logical place for this will be inside the "frontend-js-react-web" module, so we need configuration to make it easy to `import` from there, which is what this commit adds.

### DIFF
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -80,6 +80,7 @@
 				"vm-browserify": ">=0.0.4"
 			},
 			"frontend-js-react-web": {
+				"/": ">=1.0.0",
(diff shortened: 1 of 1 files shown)


jbalsas:
LGTM, hope we don't have a `destropy` method, though 😂

wincent:
I haven't actually written the wrapper yet, but [this gist](https://gist.github.com/wincent/ed73439888fbf7c56cc3da71c0f6c60e) shows where I am planning to put it.

PR #208, opened 2019-08-21
chore!: update to eslint-config-liferay v8.0.0

wincent:
Breaking change because we now prohibit the use of `render()` from "react-dom".

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"devDependencies": {
 		"eslint": "6.1.0",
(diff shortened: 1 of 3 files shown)


wincent:
Hm... minor nit I notice here; the Semantic Pull Request bot doesn't seem to know that "!" is a legitimate suffix to the "type". I'll file an upstream report about that.

wincent:
Green: https://travis-ci.org/liferay/liferay-npm-tools/jobs/574896618

PR #210, opened 2019-08-22
fix: search up as high as liferay-portal "root" for user config

wincent:
We just noticed that when you run the "format" task from a project directory, it doesn't consult the ".prettierrc.js" at the top level.

This is because we tell cosmiconfig not to walk up above the current directory. It seems reasonable that we should be able to safely walk up as high as the "root" (ie. the directory containing the "yarn.lock" file) though.

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/getMergedConfig.js
+++ b/packages/liferay-npm-scripts/src/utils/getMergedConfig.js
@@ -90,7 +90,7 @@ function getMergedConfig(type, property) {
 		case 'prettier':
 			mergedConfig = deepMerge([
 				require('../config/prettier'),
-				getUserConfig('prettier')
+				getUserConfig('prettier', {upwards: true})
 			]);
 			break;
 
--- a/packages/liferay-npm-scripts/src/utils/getUserConfig.js
+++ b/packages/liferay-npm-scripts/src/utils/getUserConfig.js
@@ -4,15 +4,17 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 
-const CWD = process.cwd();
 const cosmiconfig = require('cosmiconfig');
+const findRoot = require('./findRoot');
 
 /**
  * Helper to get configuration via `cosmiconfig`
  * @param {string} moduleName Name of user config file
  */
 module.exports = function(moduleName) {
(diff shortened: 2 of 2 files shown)


wincent:
Don't know if this is a good idea or not, so consider this an RFC.

wincent:
So on thinking about this, I don't think we want to do this unconditionally.

For example, this behavior is good for `.prettierrc.js` files, but not good for `npmscripts.config.js` (if you run `liferay-npm-scripts format` from a project directory which does not have an `npmscripts.config.js` file, you don't want it walking up and finding the top-level one, because that one [defines globs](https://github.com/liferay/liferay-portal/blob/ff3286fd16693d3660fe117d8001def232fe3c90/modules/npmscripts.config.js#L17-L19)  — eg. `'/apps/*/*/*.js'` etc — that don't make any sense in a project directory).

The other tools can pretty harmlessly walk up the tree (eg. looking for a babel or jest config that it will never find), but others might not make much sense to do (eg. ESLint).

On the balance it seems that we should do this only for Prettier (ie. whitelist) rather than for everything-except-npmscripts (ie. blacklist).

bryceosterhaus:
Seems like a good idea to me to limit it to prettier. Is there any way we can limit the depth it will search up? For example, can we specify a directory to stop at? `/liferay-portal` for example.

wincent:
> For example, can we specify a directory to stop at?

Yes, that's what this PR does (stops at the directory containing the "yarn.lock", which usually means "modules"; see the [implementation of `findRoot()`](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/utils/findRoot.js)). It doesn't literally check for "liferay-portal" because the clone of the repo doesn't necessarily need to be named that, and I'm not too worried about zipping up and out of any of the repos/branches in the liferay-portal family (because either they're on Yarn already, and so have lockfiles, or they aren't using liferay-npm-scripts yet; and even if we did go too high, the worst that would happen is we *might* find a random `.prettierrc.js` file somewhere, and that seems pretty harmless).

bryceosterhaus:
Oh nice, didn't realize it would stop at a `yarn.lock`. This looks good to me

PR #211, opened 2019-08-26
feat: provides default basic mocks for Liferay, fetch and Headers global objects

jbalsas:
Hey @wincent, this is an initial PR to get this going. It mocks all `Liferay.*` related APIs that are currently being mocked independently within portal with the exception of `Liferay.componentReady` and `Liferay.SPA` which are both being mocked with a custom implementation that keeps some local knowledge. 

After this, it should be necessary to:
- Apply to `liferay-portal`
- [Probably] Convert the `Liferay.Language.get` mock to a babel plugin
- [Maybe] Create a rule to disallow things like `Liferay = {...}` or `Liferay.foo = ...` in tests to signal that global mocks should be handled in a different way
- [Maybe] Investigate how to provide mocks that require implementations like `fetch-mock` so they could be modified locally (`.mockResponse()`).
- [As needed] Mock missing APIs


packages/liferay-npm-scripts/src/jest/mocks/liferay.js
```diff
@@ -0,0 +1,113 @@
...
+
+const authToken = 'default-mocked-auth-token';
+
+/**
+ * Event support APIs on the `Liferay` object inherited from `A.Attributes`
+ *
+ * Defines:
+ * 	- Liferay.detach
+ * 	- Liferay.fire
+ * 	- Liferay.on
```
wincent: I don't think you need this list because the defined things are visible immediately below (and the list is wrong anyway because it doesn't include `Liferay.once`).

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
```diff
@@ -0,0 +1,113 @@
...
+	/**
+	 * https://yuilibrary.com/yui/docs/api/files/event-custom_js_event-target.js.html#l136
+	 */
+	once: jest.fn()
+};
+
+/**
+ * Contains a fallback/dummy implementation of `Liferay.Language.get`. In practice, this call is rewritten in a ServerFilter, so runtime calls to
+ * `Liferay.Language.get` should not be found in production code. A better match for the real behaviour would be a babel plugin to rewrite calls
+ * to the API with their "translated" value.
```
wincent: If by this you mean replacing them with a string (not the actual string, as that seems like too much work), then I agree with you.

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * General utilities on the `Liferay`. Possible API sources are:
+ *
+ * - https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-web/src/main/resources/META-INF/resources/liferay/global.es.js
+ * - https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/util.js
+ */
+const Util = {
+	/**
+	 * https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-js/frontend-js-aui-web/src/main/resources/META-INF/resources/liferay/util.js#L442
```
wincent: All these links are doomed to become stale because what "master" points at will continuously update in the future; this is especially true of the ones that have line numbers, because the line numbers are likely to drift. (More likely than, say, a file getting renamed and a link turning into a 404).

I'd either link to a specific revision (press "y" when looking at the page you want to copy from on GitHub), or I'd drop the line numbers.

packages/liferay-npm-scripts/src/jest/mocks/headers.js
```diff
@@ -0,0 +1,47 @@
...
+
+		entries.forEach(([key, value]) => {
+			callback(value.toString(), key);
+		});
+	}
+
+	set(key, value) {
+		this.headerList[key.toLowerCase()] = [value];
+	}
+};
```
wincent: Maybe this file should be named "Headers.js", to match what it exports.

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
```diff
@@ -0,0 +1,113 @@
...
+	getPathThemeImages: jest.fn(_ => ''),
+
+	/**
+	 * https://github.com/liferay/liferay-portal/blob/master/portal-web/docroot/html/common/themes/top_js.jspf#L247
+	 */
+	getPortalURL: jest.fn(_ => 'http://localhost:8080')
+};
+
+/**
+ * General utilities on the `Liferay`. Possible API sources are:
```
wincent: Did you mean "on the `Liferay` object"?

jbalsas: I did!

packages/liferay-npm-scripts/src/jest/mocks/headers.js
jbalsas: yeah, I was wondering about that too... will rename

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
jbalsas: argh, I knew this, because you’ve told us so many times... it’s just that my muscle memory hasn’t caught up with it yet.

I’ll recreate them

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
jbalsas: yup!

packages/liferay-npm-scripts/src/jest/mocks/liferay.js
jbalsas: it’s because I added it afterwards... it proves why documentation is usually wrong 😂😂

wincent: **Annoying TypeScript Advocate:** Documentation is mostly redundant in the presence of a robust type system.

source (1)

wincent:
lg2m @jbalsas

jbalsas:
Hey @wincent, I've pushed one final mock `Liferay.Util.sub`

With this, https://github.com/jbalsas/liferay-portal/pull/1892 should be green. There are still some other mocks we could bring over that I just saw:
- `Liferay.DDM`
- `AlloyEditor`
- `Liferay.AUI`

But I think we can go over those in another pass

Issue #98, opened 2019-06-03
Restore .soy and .jsp formatting


wincent:
In #96 we replaced `csf` with `prettier` and in the process temporarily dropped support for formatting ".soy" and ".jsp" files.

That's ok as a temporary thing in the name of expediency because we have a limited window of opportunity to roll out a big change like switching to Prettier in liferay-portal. Having said that, we should go back and probably re-add "csf" just for the purposes of formatting those filetypes, because I doubt that Prettier is going to get support for those for the foreseeable future (and even if we did want to push for it and implement it ourselves, we'd need a stopgap measure).

Labelling this as a "feature", even though you could look at dropping support as a "bug" or regression. I'm just a positive guy. I can't help it.

wincent:
Created corresponding internal ticket for this here: https://issues.liferay.com/browse/LPS-98032

See also [this comment](https://github.com/brianchandotcom/liferay-portal/pull/75474#issuecomment-509839687) from Brian, which is about sorting params in soy templates — I am not sure whether `csf` was ever able to do that, but if it was, then reactivating it for ".soy" files should take care of it, and if not we'll have to take a look into how best to do it.

wincent:
Example PR that (presumably manually?) sorts params: https://github.com/brunobasto/liferay-portal/pull/1062

wincent:
Note that in https://github.com/brianchandotcom/liferay-portal/pull/77453 we discuss an alternative (or perhaps additional) approach which is to extract JS from JSP files, pipe through Prettier, and then reinject it. The obvious challenge there is making sure that all the interpolated sections don't get broken by the formatting, or in turn make the formatting produce a suboptimal result.

jbalsas:
So, the problem with good ol' `csf` is that it is a pretty manual process. The difference will still be very noticeable in jsps since they won't be applying prettier, so it will still be a very manual labour.

I think we'd better attempt leveraging Prettier somehow to make sure those stay consistent.

jbalsas:
Hey @wincent, given that we restored `.jsp` formatting and that we're mostly got rid of `.soy` in `7.2.x` where we're applying it... do you think it would make sense to close this, or do you want to keep it open as a reminder?

wincent:
Let's just close it and if we get another request to do formatting in `.soy` stuff we can create a new issue for that.

PR #214, opened 2019-09-02
feat: format JS in JSP files (#98)

wincent:
(Draft)

This PR imports the history of the check-source-formatting project ([registry page](https://www.npmjs.com/package/check-source-formatting), [repo](https://github.com/liferay/liferay-frontend-source-formatter)) so that we can format ".jsp" and maybe ".soy" files for #98.

I imported the entire history in order to preserve history and blame, then moved the files into "packages/check-source-formatting", and started trimming off bits that we obviously won't need. In the end, the bit that really matters will most likely be found in [the lib/html.js file](https://github.com/liferay/liferay-npm-tools/blob/a2b5b49b3400fc3b15d0fa9a739dadbca42a3cdf/packages/check-source-formatting/lib/html.js). At the end, the "packages/check-source-formatting" directory will be deleted; we won't actually need to publish a new version of that package.


wincent:
There might be some CSS stuff in here that we can use too, although will evaluate scss-lint or similar as well.

wincent:
Created draft PR over here to show alternative approach, which would be to build a webpack bundle of `csf`: https://github.com/liferay/liferay-frontend-source-formatter/pull/54 — Sadly, I suspect that's not going to work, or at least, not work very well, although I will return to it if I can't get the approach in this PR working properly.

wincent:
Rebase over at https://github.com/liferay/liferay-npm-tools/pull/234 where what is actually happening should be a little clearer.

PR #215, opened 2019-09-04
fix: swap order of "fix" script ("format" then "lint:fix")

wincent:
Previously we ran Prettier then ESLint in our "fix" script which meant that Prettier might wrap the code in a certain way, only to have ESLint change the AST in a way that meant that Prettier would wrap differently on the next run. We should be able to fix that by inverting the order of the fixes.

Note that because we're using `spawnMultiple` an early failure in ESLint won't stop Prettier from running (errors are accumulated and then reported in aggregate at the end).

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/fix.js
+++ b/packages/liferay-npm-scripts/src/scripts/fix.js
@@ -9,7 +9,7 @@ const format = require('./format');
 const lint = require('./lint');
 
 function fix() {
-	spawnMultiple(() => format(), () => lint({fix: true}));
+	spawnMultiple(() => lint({fix: true}), () => format());
 }
 
 module.exports = fix;


packages/liferay-npm-scripts/src/scripts/fix.js
```diff
@@ -9,7 +9,7 @@ const format = require('./format');
 const lint = require('./lint');
 
 function fix() {
-	spawnMultiple(() => format(), () => lint({fix: true}));
+	spawnMultiple(() => lint({fix: true}), () => format());
```
jbalsas: I think it won't matter, but should we do the same for the [check](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/scripts/check.js#L12) script?

wincent: Consistency is good, but the other factor here is which feedback do you think is more important? And once you know that, do you want the important feedback to appear at the beginning or the end?

jbalsas: That's an interesting thought... eslint will certainly provide more valuable feedback than prettier during the `check` phase, I guess, which better be the last information we see as it's the most costly to fix? I don't really have a strong opinion here :)

wincent: > I don't really have a strong opinion here 

Yeah, me neither. 😂 But I think you're right: ESLint provides the higher-signal feedback, so it should probably be last.

wincent:
Tested this in liferay-portal by doing a top-level `yarn format` from "modules/"; seems to work just fine.

PR #222, opened 2019-09-04
feat: add liferay-changelog-generator

wincent:
Basically took the changelog generator that we have been using in a few places (from [here](https://github.com/liferay/liferay-js-themes-toolkit/blob/640ec3bf73cc87a34eea90c4673e1881e3401be8/scripts/changelog.js))  and turned it into an NPM package. It has zero dependencies, and you can easily use it without adding it as a dependency (ie. with `npx liferay-changelog-generator`), so I hope this can make our life better without plunging us any deeper into the "Thousands-Of-Packages-Nightmare" in which we currently languish.

Once this is merged and published I'll remove it from liferay-js-themes-toolkit and update our various "CONTRIBUTING" guides to refer to this instead. There are a couple of known limitations listed in the README that I'd like to address soon too.


wincent:
"Approved via Slack", so going to merge 😂

PR #223, opened 2019-09-04
chore: remove redundant .prettierrc

wincent:
The one we have at the top-level should suffice.


wincent:
Merging because trivial.

PR #228, opened 2019-09-05
feat(changelog-generator): add short options

wincent:
Accept "-f" as a shorthand for "--force" and "-h" as a shorthand for "--help".

Test plan: Run with `--help`, `-h`, `-f`, and `--force`. Run with `f` and see it correctly report an unrecognized option.

(Just wanting to make some trivial change so that I can test the `publish.js` script with this package.)


wincent:
This package has no tests, but I did a local manual test as described in the "Test plan:", and I did a local `yarn ci` to confirm there were no formatting/lint errors.

PR #237, opened 2019-09-10
feat: updates clay v3 versions to alpha.1

jbalsas:


jbalsas:
shameless auto-merge!

bryceosterhaus:
What about projects still using a milestone, wouldn't we want to add versions instead of replace the version? These sort of updates scare me a bit since we aren't "stable" yet.

jbalsas:
This is just the optimization mechanism by which all projects share a same version. Projects not updated using a milestone (outside of DXP, I assume, because we update all those in DXP) would still use a milestone independently of this change.

bryceosterhaus:
So projects within DXP aren't able to use milestones, or are they always required to upgrade to whats specified here? I may just be a bit mistaken on this.

jbalsas:
Projects in DXP could use milestones if they choose to. By default, they share the default provided version with everyone else.

If the opt-out of the `preset-dev` they will no longer import the provided clay modules and will start bundling and serving them on their own.

By default, they're always required to upgrade. As tempting as it sounds, leaving versions uncontrolled has taken us already to an outstanding JS footprint from which we've been suffering for a long time.

bryceosterhaus:
> As tempting as it sounds, leaving versions uncontrolled has taken us already to an outstanding JS footprint from which we've been suffering for a long time.

Totally agree with you one this. This case is probably a caveat since its a pre-release. I think in the future it might be best to avoid and pre-releases in tools like this, primarily for the overhead in upgrading from one version to another. Its like that a user may have had to upgrade breaking changes from every milestone and alpha. 

We are already down this path, so i don't think its a big deal, but I think we just need to be mindful of enforcing pre-release versions through this tool.

jbalsas:
> Totally agree with you one this. This case is probably a caveat since its a pre-release. I think in the future it might be best to avoid and pre-releases in tools like this, primarily for the overhead in upgrading from one version to another. 

Yes, this has been a bit of a drag. I don't think the day we go `3.0.0` is coming fast enough 😂 

> Its like that a user may have had to upgrade breaking changes from every milestone and alpha.

We took most of this burden on ourselves. As we move onto stable and more and more teams start using Clay, we'll definitely need to take different approaches, but I think this was the most balanced approach for ourselves and DXP at this time

PR #241, opened 2019-09-11
feat: Add support for bundler css-loaders

izaera:


jbalsas:
Linux is green, so I'll go ahead and merge since this is pretty trivial.

wincent:
@izaera: This PR adds a bunch of duplicate modules to the yarn.lock; a quick scan looks like we got redundant additional versions of these packages:

- @nodelib/fs.stat
- array-union
- braces
- dir-glob
- fast-glob
- fill-range
- globby
- ignore
- is-number
- micromatch
- path-type

Just as an example, I looked at the first one (with `yarn why @nodelib/fs.stat`):

```
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "@nodelib/fs.stat"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@nodelib/fs.stat@2.0.2"
info Has been hoisted to "@nodelib/fs.stat"
   - "workspace-aggregator-b0e2b53e-16c0-485b-b11d-b0bb172b925a" depends on it
   - Hoisted from "_project_#liferay-npm-scripts#liferay-npm-bundler#globby#fast-glob#@nodelib#fs.stat"
   - Hoisted from "_project_#liferay-npm-scripts#liferay-npm-bundler#globby#fast-glob#@nodelib#fs.walk#@nodelib#fs.scandir#@nodelib#fs.stat"
info Disk size without dependencies: "56KB"
info Disk size with unique dependencies: "56KB"
info Disk size with transitive dependencies: "56KB"
info Number of shared dependencies: 0
=> Found "fast-glob#@nodelib/fs.stat@1.1.3"
info This module exists because "_project_#globby#fast-glob" depends on it.
✨  Done in 0.96s.
```

@izaera: I'm guessing that most of them are similar, coming from the bundler, so that is probably suffering from the same duplication, which we could probably eliminate. What do you think? In general, I'd like to keep our dependency footprint as small as (practically) possible.

Issue #242, opened 2019-09-12
Recover gracefully from interrupted builds


wincent:
This one was reported by @diegonvs and diagnosed by @jbalsas.

We have [a few places around where we use `finally`](https://github.com/liferay/liferay-npm-tools/search?q=finally&unscoped_q=finally) to do cleanup of temporary files we leave on the filesystem.

That deals nicely with errors but if you interrupt a build with Ctrl-C the `finally` block(s) won't run. The stale config files (eg. a merged `.babelrc`) that get left behind can cause the next run to fail with a confusing stack trace, usually deep in the bowels of the `deepMerge()` function along the lines of `target.slice() is not a function`.

We should deal with this somehow:

- Install a sig handler that does clean-up even when a build is interrupted.
- Detect a stale merged config file on disk and report the problem to the user in a way that they can understand the fix.
- Make `deepMerge()` resilient to being run on a previously merged config.

So at least one of those things and maybe two or even all of them.

Labelling this as a bug because the inability to run after an interrupted build seems broken.

PR #243, opened 2019-09-12
fix: don't choke on stale .babelrc files (#242)

wincent:
As explained in the related issue, we have a few places around where we use `finally` to do cleanup of temporary files we leave on the filesystem.

That deals nicely with errors but if you interrupt a build with Ctrl-C the `finally` block(s) won't run. The stale config files (eg. a merged `.babelrc`) that get left behind can cause the next run to fail with a confusing stack trace, usually deep in the bowels of the `deepMerge()` function along the lines of `target.slice()` is not a function.

We should deal with this in depth:

1. Make `deepMerge()` resilient to being run on a previously merged config.
2. Detect a stale merged config file on disk and report the problem to the user in a way that they can understand the fix.
3. Install a sig handler that does clean-up even when a build is interrupted.

This PR does 1, and in a way 2 (because if it has a problem merging it throws an error advising people to check for stale `.babelrc` file(s)).

I'll follow up with another PR for item 3.

Closes: https://github.com/liferay/liferay-npm-tools/issues/242

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/deepMerge.js
+++ b/packages/liferay-npm-scripts/src/utils/deepMerge.js
@@ -174,7 +174,22 @@ function babelMerge(key) {
 			);
 		};
 	} else {
-		return combineMerge;
+		return function(target, source, options) {
+			if (Array.isArray(target) && Array.isArray(source)) {
+				return combineMerge(target, source, options);
+			} else if (
+				options.isMergeableObject(target) &&
+				options.isMergeableObject(source)
+			) {
+				return merge(target, source, options);
+			} else {
+				throw new Error(
+					'babelMerge(): ' +
+						'Cannot merge without two mergeable objects: ' +
+						'please check for a stale .babelrc files'
+				);
+			}
+		};
 	}
 }
 
--- a/packages/liferay-npm-scripts/src/utils/deepMerge.js
+++ b/packages/liferay-npm-scripts/src/utils/deepMerge.js
@@ -186,7 +186,7 @@ function babelMerge(key) {
 				throw new Error(
 					'babelMerge(): ' +
 						'Cannot merge without two mergeable objects: ' +
-						'please check for a stale .babelrc files'
+						'please check that the .babelrc file is well-formed'
 				);
 			}
 		};
--- a/packages/liferay-npm-scripts/test/utils/deepMerge.js
+++ b/packages/liferay-npm-scripts/test/utils/deepMerge.js
@@ -279,6 +279,77 @@ describe('deepMerge()', () => {
 			});
 		});
 
+		it("doesn't break when a stale .babelrc file is left on disk", () => {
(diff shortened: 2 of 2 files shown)


jbalsas:
We can close this, right? this is #245?

jbalsas:
Ah, no, got it, this is just niceness on our part, #245 is just an evolution of #244  right?

packages/liferay-npm-scripts/src/utils/deepMerge.js
```diff
@@ -174,7 +174,22 @@ function babelMerge(key) {
...
+				return combineMerge(target, source, options);
+			} else if (
+				options.isMergeableObject(target) &&
+				options.isMergeableObject(source)
+			) {
+				return merge(target, source, options);
+			} else {
+				throw new Error(
+					'babelMerge(): ' +
+						'Cannot merge without two mergeable objects: ' +
```
jbalsas: One thing I was thinking is that other than stale objects on our part, developers can also mess up their config files, so this should also account for it, right?

In case this also catches that, maybe the message could also point to a syntax error in their config file?

wincent:
> Ah, no, got it, this is just niceness on our part, #245 is just an evolution of #244 right?

Correct. This one treats the symptom, and  #244 and #245 address the cause. They can all be merged independently (or not), but I did them all because each one makes us more robust. In order of importance (most to least), I'd say they are #244, #243 (ie this one) and then #245.

packages/liferay-npm-scripts/src/utils/deepMerge.js
wincent: Yeah, in theory it will catch some obviously broken config files; for example — not tested, but I think — this will blow up if you have something like this in your config file:

```json
{
    presets: null
}
```

It probably won't catch subtle cases though, and it might be tricky to provide the right guidance other than "Your config file doesn't merge cleanly: please look at it an fix it".

> a syntax error in their config file

If we're literally talking about invalid JSON, I think things already blow up long before we get this far.

jbalsas: > "Your config file doesn't merge cleanly: please look at it an fix it"

I think this plus the "it might be stale" bit should be more than enough to put them in the right track, don't you?

wincent: The thing is, thanks to the other fix (the signal handling), it is *probably* not stale. So I'll update the message to say that it needs human review, but I won't use the word "stale".

jbalsas: 👍

wincent: Revised the message text in https://github.com/liferay/liferay-npm-tools/pull/243/commits/985d9a59a71c9fcb8868df72ec8cc2aebeb3542c

PR #245, opened 2019-09-13
feat: add SignalHandler for managing cleanup

wincent:
Add a `SignalHandler` module that sets up signal handlers and provides a mechanism for registering callbacks that should be executed when a signal is received.

Originally sent this as:

https://github.com/liferay/liferay-npm-tools/pull/244

But then I realized it should be split into two bits:

1. This bit, the SignalHandler, may be overkill, and we can decide whether we want the complexity or not.
2. The other bit from #244 was making sure that `spawnSync` turns signals in child processes into errors. That on its own may be enough, and should clearly go in.

Companion to: https://github.com/liferay/liferay-npm-tools/pull/243

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/build.js
+++ b/packages/liferay-npm-scripts/src/scripts/build.js
@@ -9,6 +9,7 @@ const CWD = process.cwd();
 const fs = require('fs');
 const path = require('path');
 
+const SignalHandler = require('../utils/SignalHandler');
 const getMergedConfig = require('../utils/getMergedConfig');
 const moveToTemp = require('../utils/moveToTemp');
 const removeFromTemp = require('../utils/removeFromTemp');
@@ -42,6 +43,10 @@ function compileBabel() {
  * `liferay-npm-bundler` executable.
  */
 function runBundler() {
+	const {dispose} = SignalHandler.onExit(() => {
+		removeFromTemp('.npmbundlerrc');
+	});
+
 	try {
 		moveToTemp('.npmbundlerrc');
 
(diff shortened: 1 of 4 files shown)


jbalsas:
So, just to make sure I understand... the other bit takes care of signals within the spawned processes and this one does it for our main one so we can properly cleanup?

Only doubt I have is, should we have a common place to deal with cleanup so we don’t forget to do it in our next script, or is simple repetition good enough to detect this?

wincent:
> So, just to make sure I understand... the other bit takes care of signals within the spawned processes and this one does it for our main one so we can properly cleanup?

Correct. The reason why I described this as "overkill" is that I believe that the pattern we most commonly have is:

1. Do some work that may or may not take a while (eg. expanding file globs).
2. Write out an rc file to disk.
3. Spawn a process, which may or may not take a while (eg. Babel).
4. Clean up the rc file.

We're unlikely to leave any mess behind even without this PR because:

- Step "1" tends not to write anything go disk.
- Step "2" happens *right before* "3", any signals sent during "3" are handled by `spawnSync()`, and step "4" happens *right after*.

In other words, the odds of *actually* needing the `SignalHandler` are pretty small. My assumptions might not always be true (or remain true) though. If we feel paranoid, we ship the PR; if we reject the complexity it brings, we close it. 🤷‍♂ 

> Only doubt I have is, should we have a common place to deal with cleanup so we don’t forget to do it in our next script, or is simple repetition good enough to detect this?

Ideally, yes we would have a common place to do clean-up, but that is kind of what the `SignalHandler` is. But no matter what it is called or what it looks like, you still have to remember to "register" the cleanup that you want to get done. Even if you come up with a Grand New Abstraction that encapsulates registering cleanup and running a task in one thing, you still have to remember to use it.

jbalsas:
We’re starting to heavily use this tool, so I’d say the safer the better... I’d get this in so we don’t have to regret it later :)

Feel free to merge or wait for a second opinion if you want

PR #247, opened 2019-09-13
feat: allow for compound patterns in extensions in globs

wincent:
Back in ancient history before we knew that we were going to have to implement our own glob expansion, we made a basic `filterGlobs()` function that would take some globs like:

    ['**/*.js', '**/*.css', '**/*.foo']

And filter them down to a subset:

    ['**/*.js', '**/*.css']

To keep things simple, we didn't add support for globs that used compound patterns in their extensions (eg. `'**/*.{js,jsx}'`).

Later on, we were forced to implement our own globbing, and part of that was building a function (`preprocessGlob()`) that could preprocess complex globs (eg. `'{src,test}/**/*.js'`) and transform them into a list of simpler globs (eg. `['src/**/*.js', 'test/**/*.js']`).

Looking at the code now, I see that there's no reason to require globs to have simple extensions any more; we can support compound patterns in globs by swapping the order of operations from:

- `filterGlobs()` then `preprocessGlob()`; to:
- `preprocessGlob()` then `filterGlobs()`

I made that change in both `format.js` and `lint.js`, and then decided it was time to refactor them to extract out the common functionality into a new `getPaths()` module. The only difference between the two files in the previous implementation was that `format.js` ignored "node_modules" by default even if not explicitly configured to do so (in order to match the Prettier behavior). So, the new `getPaths()` function does this too, seeing as it is harmless to apply the same default exclusion to ESLint as well.

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/getPaths.js
+++ b/packages/liferay-npm-scripts/src/utils/getPaths.js
@@ -1,0 +1,61 @@
+/**
+ * © 2019 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const expandGlobs = require('../utils/expandGlobs');
+const filterChangedFiles = require('../utils/filterChangedFiles');
+const filterGlobs = require('../utils/filterGlobs');
+const findRoot = require('../utils/findRoot');
+const log = require('../utils/log');
+const preprocessGlob = require('../utils/preprocessGlob');
+const readIgnoreFile = require('../utils/readIgnoreFile');
+
+/**
+ * Given a set of `globs`, the file `extensions` we are interested
+ * in, and the name of an `ignoreFile` (eg. '.prettierignore' or
+ * '.eslintignore'), returns a list of matching files on the file
+ * system.
+ */
+function getPaths(globs, extensions, ignoreFile) {
+	const root = findRoot();
+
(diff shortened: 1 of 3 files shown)


packages/liferay-npm-scripts/src/utils/getPaths.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+function getPaths(globs, extensions, ignoreFile) {
+	const root = findRoot();
+
+	ignoreFile = path.join(root || '.', ignoreFile);
+
+	const ignores = fs.existsSync(ignoreFile) ? readIgnoreFile(ignoreFile) : [];
+
+	// Turn "{src,test}/*" into ["src/*", "test/*"]:
+	globs = [].concat(...globs.map(glob => preprocessGlob(glob)));
```
wincent: Once we're on a newer NodeJS version, this could use `Array.prototype.flatMap` instead.

wincent:
Testing this in liferay-portal currently. Will report back once I am sure it's good.

wincent:
> The only difference between the two files in the previous implementation was that `format.js` ignored "node_modules" by default even if not explicitly configured to do so (in order to match the Prettier behavior). So, the new `getPaths()` function does this too

Actually, this isn't true 🤦‍♂️ — I had this in my local buffer but didn't include it in the commit. Will add it.

wincent:
> Will report back once I am sure it's good.

I am 93% sure it's good.

wincent:
I guess I'll merge this then in order to keep moving forward. Will loop back and do a follow-up if there is any required.

PR #248, opened 2019-09-13
feat: format JS in JSP files using Prettier

wincent:
Before merging:

- There are some "TODO" comments to clean up.
- I'd like to do a clean-up pass, because I did a lot of iteration along the way and there are probably things which could be made nicer after looking at it all holistically.
- The version I'm pushing here initially has a secret "csf" command that I used to produce [this PR](https://github.com/wincent/liferay-portal/pull/61) in liferay-portal showing the formatting changes: once we're ready to actually ship this and run it I'll fold that into the standard "format" and "lint" subcommands.


wincent:
For some reason snapshot tests are failing on Windows: indices are off, which suggests something wrong related to Windows-style line endings (`\r\n`) vs UNIX ones (`\n`) although I have no idea why line endings are coming into it...

[Google suggests](https://github.com/nodejs/node/issues/3121) that it might be related to template literals:

> and LineTerminatorSequences are normalized to for both TV and TRV. An explicit EscapeSequence is needed to include a or sequence.

So I think the fix will be to normalize the line endings on template literals before calling `toMatchSnapshot()`.

wincent:
Fix for the Windows failures was to add a `.gitattributes` file:

```
From 5ca2873a6d9dadf8f66cd4e0b359a058a17d2515 Mon Sep 17 00:00:00 2001
From: Greg Hurrell <greg@hurrell.net>
Date: Fri, 13 Sep 2019 22:58:01 +0200
Subject: [PATCH] fix: add a .gitattributes file to unbreak Travis on Windows

My suspicion is that when Travis does its Git checkout, it converts all
LF to CRLF, and that breaks the snapshots because all the indices
change; example:

https://travis-ci.org/liferay/liferay-npm-tools/jobs/584751075

See:

- https://help.github.com/en/articles/configuring-git-to-handle-line-endings
- https://git-scm.com/docs/gitattributes
---
 .gitattributes | 6 ++++++
 1 file changed, 6 insertions(+)
 create mode 100644 .gitattributes

diff --git a/.gitattributes b/.gitattributes
new file mode 100644
index 0000000..b4195e7
--- /dev/null
+++ b/.gitattributes
@@ -0,0 +1,6 @@
+# Don't let Git transform line-endings to CRLF on Windows, because that
+# would break the snapshots.
+*.js text eol=lf
+*.jsp text eol=lf
+*.jspf text eol=lf
+*.js.snap text eol=lf
-- 
2.20.1
```

Green Windows build: https://travis-ci.org/liferay/liferay-npm-tools/jobs/584772378

wincent:
Ok, chatted with @jbalsas about this. We can merge this now because it currently doesn't do anything (because it's hidden behind a separate subcommand). Once Hugo's stuff is ready, I'll do the last step of integrating the new code transparently into existing formatting commands (ie. they'll go from checking/formatting JS/SCSS to checking/formatting JS/JSP/SCSS).

PR #253, opened 2019-09-17
feat: activate JSP formatting as part of standard formatting runs

wincent:
Note that we can drop the "simple glob" check because it is no longer required after:

https://github.com/liferay/liferay-npm-tools/pull/247

Tested here: https://github.com/wincent/liferay-portal/pull/64


packages/liferay-npm-scripts/test/format/__snapshots__/formatJSP.js.snap
```diff
@@ -928,7 +928,45 @@ exports[`formatJSP() formatting entire fixtures page.jsp matches snapshot 1`] =
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ */
+--%>
+
+<%@ include file=\\"/captcha/init.jsp\\" %>
+
+<c:if test=\\"<%= captchaEnabled %>\\">
+	<script src=\\"<%= HtmlUtil.escapeAttribute(captchaConfiguration.reCaptchaScriptURL()) %>?hl=<%= HtmlUtil.escapeAttribute(locale.getLanguage()) %>\\" type=\\"text/javascript\\"></script>
```
wincent: Note the reason this script content is included in the snapshot now is that it contains no formattable elements (it has a `<script>` tag, but it has no contents). Previously we `formatJSP` would return `null` to indicate nothing formattable, but in this PR it returns the original source, unmodified.

Same deal with the other snapshot further down.

wincent:
I think this is working well, but testing it in liferay-portal [here](https://github.com/wincent/liferay-portal/pull/64) with a prerelease (v9.5.0-beta.1) of liferay-npm-scripts. I have a suspicion that we'll need further adjustments on the Java side to actually be able to merge this, which is why I am still leaving it in "Draft" state.

wincent:
`ci:test:sf` is [green now](https://github.com/wincent/liferay-portal/pull/64#issuecomment-532877789). Still https://github.com/liferay/liferay-npm-tools/issues/259 to look at but if we want to proceed and merge this (and apply it in liferay-portal) I think we now can and do that as a fast follow-up. I'll make a quick stab at it anyway next because we might be able to sneak it in.

wincent:
Just forced push this with no changes (rebase was clean) seeing as it has been a month since this was last touched. If CI is green, will merge and test for real (again) in liferay-portal.

PR #266, opened 2019-09-20
feat: updates initial clay v3.0.0 imports

jbalsas:


bryceosterhaus:
Should we do `3.x` instead of `>=3.0.0`? Obviously we don't have 4.0 plans as of now, but it would lock it in a bit more.

jbalsas:
I guess we could, yeah. I think we can do so once we put out the final 3.x.

PR #269, opened 2019-09-23
feat: initial js-insights implementation

jbalsas:
Found some time to work on this on the plane, so I'm just dropping it here to see what we want to do with this. 

It's basically a tool that:
- **Given** a set of paths (glob)
- **Then** analyzes the code to gain insights (currently only checking external imports)
- **Then** generates a report

Different reports can be generated:

### Table (default)
report_table

### Airtable
See real [liferay-portal report](https://airtable.com/tbln9P08dd75xKAmJ/viw2InOW91BooKGuV)

report_airtable

### JSON
report_json

As I was building it, I kept thinking that we could easily add other insights like "`setInterval` usage" throughout our base code that could help us make better informed decisions.

The code is very rough... specially in the `reporters` area where I haven't really thought how to approach it. For now, reports just know what to expect from the only insight we have right now.

Anyways, creating this as a draft to see where this takes us...


yarn.lock
```diff
@@ -275,6 +275,11 @@
   resolved "https://registry.yarnpkg.com/@babel/parser/-/parser-7.4.5.tgz#04af8d5d5a2b044a2a1bffacc1e5e6673544e872"
   integrity sha512-9mUqkL1FF5T7f0WDFfAoDdiMVPWsdD1gZYzSnaXsxUCUqzuch/8of9G3VUSNiZmMBoRxT3neyVsqeiL/ZPcjew==
 
+"@babel/parser@^7.6.0":
```
jbalsas: I'm pretty sure we can fix this and use an already existing version of `@babel/parser` 😂

yarn.lock
```diff
@@ -9680,6 +9729,11 @@ lodash.values@~2.4.1:
   dependencies:
     lodash.keys "~2.4.1"
 
+lodash@4.17.14:
```
jbalsas: Not sure where this one's coming from 🙈

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+	const name = modulePath.substring(modulePath.lastIndexOf('/') + 1);
+
+	// Finds the root git folder
+	const gitRoot = await findUp('.git', {
+		cwd,
+		type: 'directory'
+	});
+
+	const url = `https://github.com/liferay/liferay-portal/blob/master/${path.relative(
```
jbalsas: Yup, this needs to be configurable :)

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+ */
+(async () => {
+	const data = [];
+
+	// eslint-disable-next-line no-for-of-loops/no-for-of-loops
+	for await (const modulePath of globby.stream(argv._[0])) {
+		const ast = await parse(modulePath);
+
+		const moduleMeta = await getModuleMeta(modulePath);
+		const moduleInfo = await INSIGHTS[0](ast, moduleMeta);
```
jbalsas: In my mind, we could loop through all available insights to get a complete picture of a module.

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+ * Considers the first non-namespaced param as the entry glob. For example:
+ *
+ * 	`npx liferay-js-insights --json modules/private/apps`
+ * 	`npx liferay-js-insights --airtable src/*.es.js`
+ */
+(async () => {
+	const data = [];
+
+	// eslint-disable-next-line no-for-of-loops/no-for-of-loops
+	for await (const modulePath of globby.stream(argv._[0])) {
```
jbalsas: Not sure if this can be done in a better way... `for await ... data.push()` 🤷‍♂

wincent:
> Found some time to work on this on the plane

Such a geek!

packages/liferay-js-insights/.yarnrc
```diff
@@ -0,0 +1,3 @@
+# Make `yarn version` produce the right commit message and tag for this package.
+version-tag-prefix "liferay-js-insights/v"
+version-git-message "chore: prepare lifeiray-js-insights/v%s"
```
wincent: lifeiray -> liferay

packages/liferay-js-insights/README.md
```diff
@@ -0,0 +1,69 @@
+# liferay-js-insights
+
+> A simple dependency analysis reporter. It scans ES6 modules for imports and generate a report on used external packages.
```
wincent: generate → generates

packages/liferay-js-insights/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Usage
+
+```
+liferay-js-insights [option...]
+
+Options:
+  --airtable    [optional: dumps the analysis to a remote airtable installation]
+  --apiKey      [optional: airtable API KEY]
+  --baseKey     [optional: airtable base]
```
wincent: I suspect all airtable-related options should be namespaced (eg. `--airtableApiKey`, `--airtableBaseKey`). You could also potentially drop the `--airtable` option then because the presence of the others could imply it?

packages/liferay-js-insights/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+By default, a table-based report will be displayed on screen unless a different reporter is selected.
+
+`npx liferay-js-insights src/**/*.es.js`
+
+![Default Table Report](/docs/img/report_table.png)
+
+### JSON Report
+
+When the `--json` flag is passed, collected insights will be dumped in a `${name}.json` file. The name of the file can be passed using the `--name` flag. If no flag is passed the name will be that of the cwd last folder.
```
wincent: Maybe when no filename is given, printing to standard-out would be a good behavior (or failing that, I should be able to make it print to standard out by specifying "-" as a filename, which is a common convention in command-line tools).

packages/liferay-js-insights/README.md
```diff
@@ -0,0 +1,69 @@
...
+- A valid API_KEY for the Airtable Base
+		- The API_KEY can be provided as `--apiKey` or as an `LFR_DEPS_AIRTABLE_API_KEY` environment variable.
+- An existing Airtable Table with the following structure:
+  - `module`:`{string}` - Name of the module with the dependencies. Acts as the primary key of the table.
+  - `app`:`{string}` - Parent application of the module.
+  - `url`:`{URL}` - GitHub URL of the module.
+  - `clay3`:`{[string]}` - List of dependencies with Clay3 packages (@clayui/*).
+  - `react`:`{[string]}` - List of dependencies with React packages (frontend-js-react-web).
+  - `js`:`{[string]}` - List of dependencies with the common package (frontend-js-web).
+  - `metal`:`{[string]}` - List of dependencies with Metl.js packages (metal-*).
```
wincent: Metl → Metal

packages/liferay-js-insights/README.md
```diff
@@ -0,0 +1,69 @@
...
+  - `clay2`:`{[string]}` - List of dependencies with Clay2 packages (clay-*).
+  - `others`:`{[string]}` - List of other dependencies.
```
wincent: I've never used Airtable but it does seem a bit odd to hard-code the major dependency names in the schema. Will it be easy to change that schema in the future when we realize we need `clay4`, `vue` 😜  etc?

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+const globby = require('globby');
+const findUp = require('find-up');
+const fs = require('fs');
+const util = require('util');
+const argv = require('minimist')(process.argv.slice(2));
+const parser = require('@babel/parser');
+const path = require('path');
```
wincent: Let me just sort these for you... I have a lint for that.

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Parses the contents of a file and returns the corresponding AST
+ */
+async function parse(path) {
+	const content = await readFile(path, 'utf8');
+
+	return parser.parse(content, {
+		plugins: ['jsx', 'classProperties'],
```
wincent: Can these be alphabetized or do they need to apply in this order?

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+		sourceType: 'module'
+	});
+}
+
+async function getModuleMeta(modulePath) {
+	// Resolves current modulePath base directory
+	const cwd = path.dirname(path.resolve(modulePath));
+
+	// Finds the closer package.json file to the modulePath
```
wincent: closer -> closest

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+	const pkg = await findUp('package.json', {cwd});
+
+	// Take app from the `name` field in package.json
+	const app = JSON.parse(await readFile(pkg, 'utf8')).name;
+
+	const name = modulePath.substring(modulePath.lastIndexOf('/') + 1);
```
wincent: Is this just the same as `path.basename(modulePath)`?

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+	// Finds the root git folder
+	const gitRoot = await findUp('.git', {
+		cwd,
+		type: 'directory'
+	});
```
wincent: There is also `git rev-pa`se --show-toplevel` which handles some edge cases that you may or may not care about like:

- if you run in a submodule (unlikely to come up for us).
- if you run in a `git worktree` checkout  (probably relevant to us).

packages/liferay-js-insights/index.js
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Considers the first non-namespaced param as the entry glob. For example:
+ *
+ * 	`npx liferay-js-insights --json modules/private/apps`
+ * 	`npx liferay-js-insights --airtable src/*.es.js`
+ */
+(async () => {
+	const data = [];
+
+	// eslint-disable-next-line no-for-of-loops/no-for-of-loops
```
wincent: I was thinking of making a "liferay/node" preset inside eslint-config-liferay that would turn off this rule for projects like this one that run on NodeJS; `for`-`of` works in pretty ancient node, and we only really have that rule because we need to do some prep to be sure that we can reliably do `for`-`of` in IE11... Or we could just turn it off in the local config.

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+	const EMPTY = [[], [], [], [], [], []];
+	const SOURCE_GROUPS = [
+		source => source.includes('@clayui'),
+		source =>
+			source.includes('react') ||
+			source.includes('prop-types') ||
+			source.includes('classnames'),
+		source => source.includes('frontend-js-web'),
+		source => source.includes('metal-') || source === 'metal',
+		source => source.includes('clay-') || source === 'clay',
```
wincent: Shouldn't all these `includes` be `startsWith`?

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+	];
+
+	function getGroup(source) {
+		return SOURCE_GROUPS.findIndex(group => group(source));
+	}
+
+	const externalImports = ast.program.body.filter(
+		node =>
+			node.type === 'ImportDeclaration' &&
+			!node.source.value.includes('./')
```
wincent: I think you want `startsWith` here too, but seeing as you also need to check fo `'../'` (and arguably leading `/` as well, although I have a lint to stop those) maybe you should just use a regex. It could be something as simple as `/^[a-z@]/i.test(node.source.value)`.

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+ * - Imports are grouped per origin into:
+ * 	- clay3
+ *  - react
+ *  - js
+ *  - metal
+ *  - clay2
+ *  - others
+ */
+function extractImports(ast) {
+	const EMPTY = [[], [], [], [], [], []];
```
wincent: This is a bit fragile because you have to keep it in sync with the list below. You could instead create it based on the `length` of `SOURCE_GROUPS`.

```javascript
const EMPTY = Array.from(new Array(SOURCE_GROUPS.length), () => []);
```

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+
+			imports[group] = [
+				...imports[group],
+				node.specifiers
+					.filter(specifier => specifier.local.name)
+					.map(specifier => specifier.local.name)
+			].reduce((memo, it) => memo.concat(it), []); // Not using flatMap to support Node.js 10;;;
+
+			return imports;
+		}, EMPTY) || EMPTY)
```
wincent: Why do we need this `|| EMPTY` here? `reduce` will return `EMPTY` for you anyway.

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+	];
+}
+
+/**
+ *
+ */
```
wincent: Let's put something in the comment or delete it.

packages/liferay-js-insights/insights/dependencies.js
```diff
@@ -0,0 +1,80 @@
...
+module.exports = async function(ast, insights) {
+	const [count, clay3, react, js, metal, clay2, others] = extractImports(ast);
```
wincent: This is also order-sensitive, which seems a bit brittle, especially because it is far away from the `SOURCE_GROUPS` declaration. Did you think about using an object or `Map` instead?

packages/liferay-js-insights/reporters/airtable.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * 		- The Table name should be provided as `--name`. If none is passed, it defaults to `master`.
+ */
+module.exports = async function(modulesInfo, config) {
+	const {apiKey, baseKey, table} = {
+		...{
+			apiKey: process.env.LFR_DEPS_AIRTABLE_API_KEY,
+			baseKey: process.env.LFR_DEPS_AIRTABLE_BASE_KEY,
+			table: 'master'
+		},
```
wincent: You don't seem to need the `...{`/`}` wrapping here: just set the keys.

packages/liferay-js-insights/reporters/airtable.js
```diff
@@ -0,0 +1,82 @@
...
+		...config
```
wincent: Looks like `argv` will override `env` here. Is that you intent? (Just double-checking. I guess that's the right thing to do).

packages/liferay-js-insights/reporters/airtable.js
```diff
@@ -0,0 +1,82 @@
...
+			});
+		} catch (err) {
+			console.error(err);
+		}
+
+		synced += 10;
+		progressBar.update(synced);
+	}
+
+	progress.stop();
```
wincent: Might be nice to print a summary at the end (ie. synced count  & error count?)

packages/liferay-js-insights/reporters/json.js
```diff
@@ -0,0 +1,49 @@
...
+ * Filters unwanted dependencies from a module's metadata to simplify the final report.
+ */
+function prepareModule(dependencies) {
+	const simpleModule = {};
+
+	Object.keys(dependencies)
+		.filter(key => dependencies[key] && dependencies[key].length)
+		.forEach(key => {
+			simpleModule[key] = dependencies[key];
+		});
```
wincent: Pointless micro-optimization; you could do one pass through the keys instead of two, and seeing as you are already using `for`/`of`, we could also just turn off that lint rule in this project and do:

```javascript
for ([key, value] of Object.entries(dependencies)) {
  if (value && value.length) {
    simpleModule[key] = value;
  }
}
```

Don't know if that is more or less readable though. 🤷‍♂

packages/liferay-js-insights/reporters/table.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+const Table = require('cli-table');
+
+/**
+ * Prints the modules's insights report as a table in the terminal. By default, it only includes the `app`, `module`, `clay3` and `react` fields for brevity.
+ */
+module.exports = async function(modulesInfo) {
+	const table = new Table({
+		colWidths: [35, 35, 50, 70],
```
wincent: Are these suggested widths or hard min/max limits? Just asking because we could probably calculate optimal widths if we wanted to avoid wasted space.

yarn.lock
wincent: I'm just going to pretend I didn't see this file at all.

jbalsas: 👀

packages/liferay-js-insights/README.md
jbalsas: 👍

packages/liferay-js-insights/README.md
jbalsas: So, you would also expect to be able to see the json report in a terminal? We'd need to pretty-print it then :)

packages/liferay-js-insights/README.md
jbalsas: >  it does seem a bit odd to hard-code the major dependency names in the schema

It does

> Will it be easy to change that schema in the future when we realize we need clay4, vue 😜 etc?

It won't 😂 

Airtable doesn't offer any meta APIs, so there's no way to get a table's schema programatically. We could probably abstract this a bit to take some type of manually generated schema (`schema.js`) that could both configure the fields and how to retrieve them from the imports...

packages/liferay-js-insights/index.js
jbalsas: 😁

packages/liferay-js-insights/index.js
jbalsas: Probably :)

Actually, I had to add `classProperties` because it's still present in some code in `segments-web`. Didn't we forbid that? Is it enabled by default in our current version in `npm-scritps`?

packages/liferay-js-insights/index.js
jbalsas: I'll take a look! 👍

packages/liferay-js-insights/index.js
jbalsas: sounds good to me!

packages/liferay-js-insights/insights/dependencies.js
jbalsas: yeah, probably... wondering if I did that on purpose or just copy-pasting...

packages/liferay-js-insights/insights/dependencies.js
jbalsas: 👍

packages/liferay-js-insights/insights/dependencies.js
jbalsas: I don't remember! 😂

packages/liferay-js-insights/insights/dependencies.js
jbalsas: NO! That comment says exactly what it means to say!

packages/liferay-js-insights/insights/dependencies.js
jbalsas: I already forgot why I did it this way... probably had a bad reason for it 😂 

I'll probably take a look on the plane back :D

packages/liferay-js-insights/reporters/json.js
jbalsas: Yeah, I've heard of people actually wanting to ban the use of `if`, so I was trying to avoid it :D 

I like how `.filter` reads, and of course, `keys` here is but a minuscule set. On the other hand, we're already mutating something in a `forEach`, so... 🤷‍♂

packages/liferay-js-insights/reporters/table.js
jbalsas: Just manual trial and error on my screen. And we all know which screen matters here ;)

jbalsas:
Hey @wincent, thanks for the quick review! I'll see if I can put some additional time on this in the next few days.

My biggest doubt is... do you think there's value in us having something like this? Do you think we'll use it long-term, or it's just a summer-fling? Don't want us spending a lot of time on this if we don't really have use for it... 🤔

packages/liferay-js-insights/README.md
wincent: TBH, I wasn't thinking about eyeballing it but rather piping it into another tool. Making your command-line tools pipe-friendly is a good practice.

But yeah, if you want to make it pretty, a `JSON.stringify(..., null, 2)` wouldn't hurt.

jbalsas: Makes sense, gotta love piping!

packages/liferay-js-insights/index.js
wincent: It is enabled for everybody: https://github.com/liferay/liferay-npm-tools/blob/2d2f90a660b94a4b855eb7cb4912930be675146a/packages/liferay-npm-scripts/src/config/babel.json#L4

packages/liferay-js-insights/insights/dependencies.js
wincent: giphy-downsized-large

packages/liferay-js-insights/reporters/json.js
wincent: > I've heard of people actually wanting to ban the use of `if`, so I was trying to avoid it 

Radical FP ~enthusiasts~ theoreticians.

wincent:
> My biggest doubt is... do you think there's value in us having something like this? Do you think we'll use it long-term, or it's just a summer-fling? Don't want us spending a lot of time on this if we don't really have use for it...

Well, it is fairly self-contained so we can always use it for a bit (and to be honest, it is probably going to be all you that does the work!) and get rid of it later if we see that it isn't providing enough value.

On the other hand, if you want to register a domain name for it and make a marketing site, then we should maybe have a chat. 😂

jbalsas:
> On the other hand, if you want to register a domain name for it and make a marketing site, then we should maybe have a chat. 😂

You know me so well already! 😍

packages/liferay-js-insights/reporters/airtable.js
jbalsas: Yup, that's intentional, take defaults from here but override with user-provided defaults.

jbalsas:
Hey @wincent, I've applied most of your suggested changes. Could you (or anyone) take another look to see if we should land this?

Still a lot of work to be done, but this does the trick for now and maybe we'll never touch it or use it again :D

jbalsas:
Pushed some additional fixes/features. Will resend to take this out of draft

jbalsas:
See https://github.com/liferay/liferay-npm-tools/pull/279

PR #271, opened 2019-09-23
fix(publish): loosen upstream remote regex

wincent:
When testing this with eslint-config-liferay, discovered that we were being juuuuust a little bit too strict in here.

This won't catch repos under the Metal org, but basically everything that we're actively developing is under "liferay/" so this is probably good.


wincent:
Going to merge this and cut a release so I can test it over here: https://github.com/liferay/eslint-config-liferay/pull/93

wincent:
Green: https://travis-ci.org/liferay/liferay-npm-tools/jobs/588446750

PR #280, opened 2019-10-02
fix(scripts): make soy.js correctly expand globs on Windows

wincent:
Addresses the failure to translate soy files reported here:

https://github.com/petershin/liferay-portal/pull/724#issuecomment-537253992


jbalsas:
So close and so far at the same time!! 😂

wincent:
In general we need to a lot more test coverage in liferay-npm-scripts to make sure that we don't break anything on windows. We are splitting, joining, parsing, constructing paths all over the place, and passing them into various places (`fs` API, webpack, Babel, Jest etc). It's all pretty scary.

PR #283, opened 2019-10-08
feat: cleans up rogue sourcemaps inside liferay-npm-bridge-generator output

jbalsas:


packages/liferay-npm-scripts/src/scripts/build.js
```diff
@@ -62,11 +62,44 @@ function runBundler() {
...
+// To ensure output is the same after 1st and 2nd pass, we need to cleanup rogue
+// sourcemaps in the bridge-generator output folder.
+//
+// See https://github.com/petershin/liferay-portal/pull/724 for more details.
+//
 function runBridge() {
 	spawnSync('liferay-npm-bridge-generator');
+
+	// Retrieves the `.npmbridgerc` configuration which we already know exists
+	const bridgeConfig = JSON.parse(
```
jbalsas: Not sure if we need to try-catch this or we have some error boundaries in place already?

wincent: I am not sure, but I would guess that if it is not valid JSON the bridge generator itself may already have exploded (and if not, it should have produced something totally broken, which would probably get noticed pretty quickly). Either way, exploding here for invalid JSON seems appropriate.

packages/liferay-npm-scripts/src/scripts/build.js
```diff
@@ -62,11 +62,44 @@ function runBundler() {
...
+		fs.readFileSync(path.join(CWD, '.npmbridgerc'), 'utf8')
```
wincent: I believe at this point you're still in the same directory (we don't change directories without changing back; AFAIK, anywhere we change directory is guarded by a `try`/`finally` block), so I suspect this could be:

```javascript
fs.readFileSync('.npmbridgerc', 'utf8')
```

packages/liferay-npm-scripts/src/scripts/build.js
```diff
@@ -62,11 +62,44 @@ function runBundler() {
...
+	);
+
+	Object.keys(bridgeConfig).forEach(moduleKey => {
+		const output = bridgeConfig[moduleKey].output;
+
+		if (output) {
+			spawnSync('rimraf', [`${output}/**/*.map`]);
```
wincent: I don't know why we're using `spawnSync` here instead of calling the `rimraf.sync` API directly (I know the one other place in this project that we use `rimraf`, we do it in the same way, but I am not sure there is a good reason for it).

packages/liferay-npm-scripts/src/scripts/build.js
```diff
@@ -62,11 +62,44 @@ function runBundler() {
...
+// 1st pass
+// - runBundler
+// - runBridge: will expand output with an additional `bridge` folder
+//
+// 2nd pass
+// - runBundler: will now see `bridge` folder to process generating sourcemaps
+// - runBridge: depending on the configuration it can re-process previous `bridge`
+// 				folders, so it's advised to always add `!**/bridge` inside the
+//				`.npmbridgerc` `file-globs` configuration.
+//
```
wincent: Just to make this super clear, I think I'd actually include some sample paths:

```
// In our common workflow right now, runBundler and runBridge can interact to
// create some kind of circular dependency where files keep being re-processed
// over and over or are only processed after the first pass.
//
// Given a typical config, where the bundler is configured to write output to "build/..."
// and the bridge generator is configured to read from "build/..." and write to
// "build/.../bridge":
//
// 1st pass
// - runBundler: operates on output folder (eg. "build/...")
// - runBridge: will expand output with an additional `bridge` folder (eg. "build/.../bridge")
```

jbalsas:
I've tested and the `.npmbridgerc` change and push-forced that together with the additional comments.

I've created https://github.com/liferay/liferay-npm-tools/issues/284 for the `rimraf` issue.

PR #286, opened 2019-10-08
feat(scripts): check bundled Babel config file for validity

wincent:
Make sure the kind of mistake (a non-canonical plugin or preset name) like the one fixed in:

https://github.com/liferay/liferay-npm-tools/pull/285

doesn't happen again.


wincent:
Pretty trivial, so going to merge this.

PR #287, opened 2019-10-09
fix(scripts): disables useStrict in commonjs transform by default

jbalsas:
/cc @antonio-ortega

### DIFF
--- a/packages/liferay-npm-scripts/src/config/babel.json
+++ b/packages/liferay-npm-scripts/src/config/babel.json
@@ -3,7 +3,13 @@
 	"plugins": [
 		"@babel/proposal-class-properties",
 		"@babel/proposal-export-namespace-from",
-		"@babel/proposal-object-rest-spread"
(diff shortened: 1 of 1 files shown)


packages/liferay-npm-scripts/src/config/babel.json
```diff
@@ -3,7 +3,13 @@
 	"plugins": [
 		"@babel/proposal-class-properties",
 		"@babel/proposal-export-namespace-from",
-		"@babel/proposal-object-rest-spread"
+		"@babel/proposal-object-rest-spread",
+		[
+			"@babel/transform-modules-commonjs",
+			{
+				"strictMode": false
```
wincent: In case anybody is wondering where this option is documented it's in https://babeljs.io/docs/en/babel-plugin-transform-strict-mode/ and not in https://babeljs.io/docs/en/next/babel-plugin-transform-modules-commonjs.html as you might expect.

PR #289, opened 2019-10-10
feat(preset-liferay-dev): updates supported @clayui imports versions to 3.0.0

jbalsas:


packages/liferay-npm-bundler-preset-liferay-dev/config.json
```diff
@@ -113,36 +113,38 @@
...
 				"@clayui/popover": ">=3.0.0",
 				"@clayui/progress-bar": ">=3.0.0",
 				"@clayui/shared": ">=3.0.0",
 				"@clayui/slider": ">=3.0.0",
 				"@clayui/sticker": ">=3.0.0",
-				"@clayui/table": ">=3.0.0-alpha.1",
+				"@clayui/table": ">=3.0.0",
 				"@clayui/tabs": ">=3.0.0",
-				"@clayui/time-picker": ">=3.0.0-alpha.1",
+				"@clayui/time-picker": ">=3.0.0",
```
jbalsas: Hey @bryceosterhaus, just in case you're wondering, I haven't change the `>=` yet because I was debating on wether we'd just want something like `^3.0.0` or `(3.0.0,${last_integrated}]` to allow teams using higher versions locally... this last option is currently not allowed by SF and getting it to work (if we wanted) might be challenging but I didn't want to simply discard the option.

Just some background 😉 

I was thinking that we might want to go with something like: `(3.0.0,${last_integrated}]` where `last_integrated` would be the last version of the modules that were updated and are being provided at runtime in DXP.

Imagine we have for `@clayui/nav` the following:
- last published: 3.4.6
- version in dxp: 3.4.5

In the current setup, all packages need to use `3.4.5` by default, enforced by SF.

If a specific project disabled that rule and asked for `3.4.6` to get a released package, chances are they would still be served `3.4.5` at runtime because we're sa

PR #290, opened 2019-10-10
feat(scripts): update preset to latest (v2.1.0)

wincent:
Labeling this as a feature because the new version of the preset makes new Clay modules available.

Note that I'm sending this against "master" but we have to cut a maintenance release that is basically "master" minus the changes to the build output directory (d539eca etc). I'll prepare that after this and will see which way looks less hideous; either some revert commits on master or another branch.

### DIFF
--- a/packages/liferay-npm-scripts/package.json
+++ b/packages/liferay-npm-scripts/package.json
@@ -46,7 +46,7 @@
 		"liferay-lang-key-dev-loader": "^1.0.3",
 		"liferay-npm-bridge-generator": "2.13.3",
 		"liferay-npm-bundler": "2.13.3",
-		"liferay-npm-bundler-preset-liferay-dev": "2.0.1",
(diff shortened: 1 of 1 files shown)


wincent:
Looks like the optimal plan will be:

1. Merge this so that "master" reflects what we want it to look like.
2. Make a temporary branch so that we can prepare a temporary revert of d539ecacf92a06 (so that we can go to liferay-portal).
    - That in turn requires us to cut releases of the preset and the scripts.
    - We keep the "master" branch in its current form and cut "beta" releases from it if Peter Shin needs to do any further testing; once he's given us the green light, we cut a non-beta release from master.
    - If we need to cut other immediate releases, we do that off the temporary branch.
3. When that's all done, we can actually delete the temporary branch so it won't clutter up the repo (the commits will be preserved and reachable because they'll be tagged).

PR #296, opened 2019-10-22
chore(scripts)! update eslint-config-liferay to v13.0.0

wincent:
Release notes:

https://github.com/liferay/eslint-config-liferay/releases/tag/v13.0.0

Note the breaking change there, which is that we have to explicitly configure the "notice/notice" rule in order to make per-directory overrides actually possible.


wincent:
Green: https://travis-ci.org/liferay/liferay-npm-tools/jobs/601226009

PR #297, opened 2019-10-22
feat: prevent accidental use of top-level `yarn version`

wincent:
So that if you accidentally do `yarn version --major` at the top level, we'll fail fast instead of bumping the version number and creating a useless commit and tag.


package.json
```diff
@@ -16,6 +16,7 @@
 		"format:check": "prettier --list-different '**/*.js' '**/*.json' '**/*.md'",
 		"lint": "eslint '.*.js' '**/*.js'",
 		"lint:fix": "eslint --fix '.*.js' '**/*.js'",
+		"preversion": "echo Cannot version top-level private package; false",
```
bryceosterhaus: this is a great idea, never thought about doing this. 🥇

PR #305, opened 2019-11-27
feat(preset): add frontend-js-components-web to the preset

wincent:
This is a new module where we will be putting React components that don't belong in Clay itself, but which are used in multiple places throughout liferay-portal.

We not putting them in frontend-js-web to avoid bloat there, nor in frontend-js-react-web because that's just plumbing for using React in liferay-portal. You could think of these as being analogous or related to the Clay "Satellites" defined in Lexicon.


wincent:
Got the over-the-shoulder review from @jbalsas on this one, so I am going to merge it.

PR #308, opened 2019-11-27
feat(scripts): teach Jest to deal with new liferay-npm-bundler loaders

wincent:
As explained in the inline comments, these transforms allow Jest to work with components that contain SCSS and CSS imports:

     import './MyComponents.scss';

Given such an import Jest will otherwise try to read the file and treat it as JS, which obviously won't work. We use the transform here to turn such "modules" into harmless no-ops.


wincent:
As I mentioned in Slack, this one is pretty benign and liferay-portal tests all still pass so I think I'll merge it and we can address any feedback in a follow-up; some sample test output:

liferay-portal___git

PR #311, opened 2019-11-29
chore(scripts): prepare liferay-npm-scripts v15.0.0-beta.3 release

wincent:
Testing this out in liferay-portal so we can show the results of the Prettier post-processing to Brian.

Just a draft, not for "official use"; no need to merge.


wincent:
Closing this because it was only ever here for visibility. If we end up shipping the related feature I will have to cut a new release anyway.

PR #312, opened 2019-12-02
docs(scripts): remove stale link from README

wincent:
Stale because we aren't using the `check-source-formatting` package
anymore.


wincent:
Trivial so merging.

Issue #309, opened 2019-11-27
Activate linting inside JSP files


wincent:
A subset of ESLint lints, because there are lots of things inside JSP that we can't meaningfully check due to lack of context. But this will be good because it means we can catch things like use of `const` etc that might otherwise slip through.

PR #313, opened 2019-12-02
refactor(scripts): rename src/format to src/jsp

wincent:
Because soon — in service of https://github.com/liferay/liferay-npm-tools/issues/309 — the code in this directory will be doing more that just formatting JSP; it will also be linting it. Let's rename the directory in order to avoid confusion.


wincent:
CI is green and this is a straight rename so going to merge and continue building on this.

PR #316, opened 2019-12-04
feat: don't create temp files in src directory

izaera:


packages/liferay-npm-scripts/package.json
```diff
@@ -39,13 +39,14 @@
 		"deepmerge": "^4.0.0",
 		"eslint": "6.5.1",
 		"eslint-config-liferay": "13.0.0",
+		"fs-extra": "^8.1.0",
```
wincent: 🤦‍♂️ we now have four copies of this dependency in this project alone:

```
$ yarn why fs-extra
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "fs-extra"...?
=> Found "fs-extra@8.1.0"
info Has been hoisted to "fs-extra"
   - "workspace-aggregator-7e6f7756-ea7d-478a-b6cb-391cbafc5d06" depends on it
   - Hoisted from "_project_#liferay-npm-scripts#fs-extra"
   - Hoisted from "_project_#liferay-npm-scripts#liferay-npm-bridge-generator#fs-extra"
   - Hoisted from "_project_#liferay-npm-bundler-preset-liferay-dev#liferay-npm-bundler-plugin-replace-browser-modules#fs-extra"
   - Hoisted from "_project_#liferay-npm-scripts#liferay-npm-bundler#fs-extra"
=> Found "liferay-theme-tasks#fs-extra@0.24.0"
info This module exists because "_project_#liferay-npm-scripts#liferay-theme-tasks" depends on it.
=> Found "@storybook/core#fs-extra@8.0.1"
info This module exists because "_project_#liferay-npm-scripts#@storybook#react#@storybook#core" depends on it.
=> Found "file-system-cache#fs-extra@0.30.0"
info This module exists because "_project_#liferay-npm-scripts#@storybook#react#@storybook#core#file-system-cache" depends on it.
```

liferay-portal (without this PR) has 6:

```
=> Found "fs-extra@0.24.0"
=> Found "liferay-npm-bundler#fs-extra@4.0.3"
=> Found "liferay-npm-bridge-generator#fs-extra@5.0.0"
=> Found "liferay-npm-bundler-plugin-replace-browser-modules#fs-extra@7.0.1"
=> Found "@storybook/core#fs-extra@8.1.0"
=> Found "file-system-cache#fs-extra@0.30.0"
```

packages/liferay-npm-scripts/src/presets/standard/index.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
 	build: {
 		dependencies: [...clay, ...liferay, ...metal],
 		input: 'src/main/resources/META-INF/resources',
-		output: 'build/node/packageRunBuild/resources'
+		output: 'build/node/packageRunBuild/resources',
+		temp: 'build/npmscripts'
```
wincent: Are there any special considerations with adding a new subfolder under `build/`? ie. is there any tooling anywhere else that might be affected? Also, wondering why this isn't under `build/node`, seeing as it is part of the "node build" (although I don't think we can push it further down under `build/node/packageRunBuild` because I suspect our Gradle SDK ends up just copying everything out of there).

packages/liferay-npm-scripts/src/scripts/test.js
```diff
@@ -51,10 +50,8 @@ module.exports = function(arrArgs = []) {
 			);
 		}
 
-		withBabelConfig(() => {
-			spawnSync('jest', ['--config', CONFIG_PATH, ...arrArgs.slice(1)], {
-				env
-			});
+		spawnSync('jest', ['--config', CONFIG_PATH, ...arrArgs.slice(1)], {
+			env
```
wincent: Is this change legit? I would expect that we probably have tests in liferay-portal that won't work properly without our standard Babel config as a base.

packages/liferay-npm-scripts/src/scripts/webpack.js
```diff
@@ -33,13 +31,30 @@ module.exports = function(...args) {
...
 	} else {
-		spawn('webpack', args);
+		withWebpackConfig(WEBPACK_CONFIG_FILE, configFilePath => {
+			spawnSync('webpack', ['--config', configFilePath, ...args]);
+		});
 	}
 };
+
+function withWebpackConfig(filename, callback) {
+	const mergeBabelLoaderOptionsPath = require.resolve(
```
wincent: This should also be `path.resolve`, right? You've got a relative path and you don't need to involve the Node dependency resolution algorithm.

packages/liferay-npm-scripts/src/scripts/webpack.js
```diff
@@ -33,13 +31,30 @@ module.exports = function(...args) {
...
+		'../utils/mergeBabelLoaderOptions'
+	);
+	const webpackConfigPath = path.resolve(filename);
+
+	const webpackConfig = `
+module.exports = require('${mergeBabelLoaderOptionsPath}')(require('${webpackConfigPath}'));
+`;
```
wincent: May as well make this look "pretty":

```javascript
    const webpackConfig = `
        module.exports = require('${mergeBabelLoaderOptionsPath}')(
            require('${webpackConfigPath}')
        );
    `;
```

packages/liferay-npm-scripts/src/utils/soy.js
```diff
@@ -4,35 +4,42 @@
...
+		path.join(BUILD_CONFIG.input, '**', '*.soy'),
+		'-d',
+		SOY_TEMP_DIR,
 		'--soyDeps',
 		generateSoyDependencies(BUILD_CONFIG.dependencies),
 		'--externalMsgFormat',
 		"Liferay.Language.get('$2')"
 	]);
+
+	runBabel(SOY_TEMP_DIR, '--out-dir', BUILD_CONFIG.output, '--source-maps');
```
wincent: Does Babel do anything useful for us here? If I understand correctly, before this PR, we used to run:

1. Compile soy (in place)
2. Babel
3. Bundler

So, we were using Babel on the compiled soy before, but I don't know what is in there that would need transforming with Babel (the bit we do need — wrapping in `Liferay.Loader.define` etc — happens in step 3 anyway).

packages/liferay-npm-scripts/src/utils/withTempFile.js
```diff
@@ -0,0 +1,52 @@
...
+ * assumed to be synchronous).
+ *
+ * @param {string} filename
+ * The name of the file.
+ *
+ * @param {string|Buffer} content
+ * Content of the file.
+ *
+ * Note that the content is used as a seed to derive an HEX string to be used as
+ * suffix to avoid collisions.
```
wincent: This comment seems untrue.

packages/liferay-npm-scripts/test/scripts/webpack.js
```diff
@@ -31,31 +31,51 @@ describe('scripts/webpack.js', () => {
...
 	it('invokes webpack', () => {
 		webpack();
-		expect(spawnSync).toHaveBeenCalledWith('webpack', []);
+
+		expect(spawnSync.mock.calls).toHaveLength(1);
+		expect(spawnSync.mock.calls[0]).toHaveLength(2);
+		expect(spawnSync.mock.calls[0][0]).toBe('webpack');
+		expect(spawnSync.mock.calls[0][1]).toHaveLength(2);
+		expect(spawnSync.mock.calls[0][1][0]).toBe('--config');
+		expect(spawnSync.mock.calls[0][1][1]).toMatch(/webpack.config.js$/);
```
wincent: All of this expansion seems like overkill that harms readability/writability to guard against a bug that will never happen. So you know there was exactly one call now, but do you *really* think that we'd accidentally call it twice?

packages/liferay-npm-scripts/src/scripts/webpack.js
izaera: Mmm, I think it should be `require.resolve`. I'm getting the absolute path of the file that corresponds to a node module so that I can then inject it in the temporary `webpack.config.js` file that will be read and loaded by webpack in a different process. 

I could also compute it with `path.join` and `__dirname` since I'm in a sibling module, but `require.resolve` looks more elegant, IMO...

packages/liferay-npm-scripts/package.json
izaera: What do you suggest? That's always going to happen as there are projects we don't have control over.

Maybe we can use the same version for our projects but again it's going to be difficult to make them converge always. It can even lead to needing spurious releases just to make sure all our projects share the same version...

However, it's strange that `liferay-npm-bundler`, `liferay-npm-bridge-generator` and `liferay-npm-bundler-plugin-replace-browser-modules` use different versions since I unified that 1-2 months ago. It is a smell. Maybe the portal is using more than one version of the JS Toolkit too and that's the root cause.

packages/liferay-npm-scripts/src/presets/standard/index.js
izaera: I didn't want to put it under `node` because that is controlled by our gradle plugin and it used to put more things (in addition to `packageRunBuild`) so I wanted to minimize the possibility of a conflict.

There's no standard about what may be put in `build`, but it looks reasonable to have one folder per _logic unit_ so that they don't collide.

packages/liferay-npm-scripts/src/scripts/test.js
izaera: It probably breaks everything :muscle: I wanted to get rid if `withBabelConfig` since it is used only there, but seeing that we need it, I may move it to that module, for example...

I don't like the idea of leaving a `withBabelConfig` hanging around in its own module since someone may use it for build purposes and break `deployFast` again... What do you think?

packages/liferay-npm-scripts/src/utils/soy.js
izaera: The soy compiler generates `import from...` statements, that's why it needs transpiling. :woman_shrugging:

packages/liferay-npm-scripts/src/utils/withTempFile.js
izaera: Yes. It's a leftover of a previous version :sweat_smile:

packages/liferay-npm-scripts/test/scripts/webpack.js
izaera: We can remove the `toHaveLength` expectations, but the problem was with the last argument (the `toMatch` one) that cannot be tested with `toHaveBeenCalledWith` because it is an absolute path into the developer's machine :-(.

If you can think of any other way to test this with a cleaner syntax I would be glad to change it because it looks horrible to me too :sweat_smile: .

packages/liferay-npm-scripts/package.json
wincent: > What do you suggest? That's always going to happen as there are projects we don't have control over.

Nothing sadly, short of avoiding using dependencies when it is practical to do so. I don't even know what you're using from `fs-extra` here; is it just `ensureDirSync` or something else as well?

packages/liferay-npm-scripts/src/scripts/test.js
wincent: Maybe just do it locally here without re-instating `withBabelConfig`. That is probably better than making it `withBabelConfig__DO_NOT_USE_OR_YOU_WILL_BE_FIRED`.

packages/liferay-npm-scripts/src/scripts/webpack.js
wincent: I *guess* `require.resolve` will short-circuit as soon as it sees you using a relative path anyway, so it won't invoke the entire machinery of the node resolution algorithm ie. it will be basically the same as what I was suggesting you write.

packages/liferay-npm-scripts/src/utils/soy.js
wincent: 😱

packages/liferay-npm-scripts/test/scripts/webpack.js
wincent: I think you can test it with `toHaveBeenCalledWith`:

```javascript
expect(spawnSync).toHaveBeenCalledWith(
    'webpack', [
        '--config',
        expect.stringMatching(/\bwebpack\\.config\\.js$/,        
    ]
);	
```

There's also [`expect.stringContaining`](https://jestjs.io/docs/en/expect#expectnotstringcontainingstring) if you don't want the rigor of the regular expression.

packages/liferay-npm-scripts/package.json
izaera: I'm using `removeDirSync` instead of  `rimraf` to get rid of the whole directory tree. We may as well leave the folder tree empty, but I preferred to delete everything to make sure leftovers don't interact with the build.

izaera: > Nothing sadly, 

:cry:

packages/liferay-npm-scripts/test/scripts/webpack.js
izaera: :-o :clap:

Issue #317, opened 2019-12-04
Don't print ESLint messages in red if there are no errors or warnings


wincent:
Textual output:

```
$ liferay-npm-scripts fix
✖ 0 problems (0 errors, 0 warnings)
Prettier checked 11 files, fixed 1 file
✨  Done in 1.99s.
```

Screenshot of output showing unwanted red text:

![liferay-portal___portool](https://user-images.githubusercontent.com/7074/70162853-aff6b600-16be-11ea-9d10-f4e8764ed0a4.png)

PR #318, opened 2019-12-04
fix(scripts): don't use color or report results if successful (#317)

wincent:
Only report if there are errors or warnings, and do so in red for errors and yellow if there are only warnings. This matches the ESLint behavior.

Closes: https://github.com/liferay/liferay-npm-tools/issues/317


wincent:
This is pretty simple (no core logic changes, just output format) so going to merge it.

PR #328, opened 2020-01-08
fix: make windows build work

izaera:


packages/liferay-npm-scripts/src/scripts/webpack.js
```diff
@@ -47,10 +47,18 @@ module.exports = function(...args) {
...
 
 function withWebpackConfig(filename, callback) {
-	const mergeBabelLoaderOptionsPath = require.resolve(
+	let mergeBabelLoaderOptionsPath = require.resolve(
 		'../utils/mergeBabelLoaderOptions'
 	);
-	const webpackConfigPath = path.resolve(filename);
+	let webpackConfigPath = path.resolve(filename);
+
+	if (path.sep === '\\') {
```
jbalsas: Isn't there a better check to do than `path.sep` to distinguish the current environment?

jbalsas: Do we actually need to even check? Would a blind replacement affect *nix environments?

izaera: Well, it's what we are replacing, the `path.sep` itself. Strictly talking, it has nothing to do with Windows vs Linux.

izaera: And to the second question: yes, you are right. I thought it would break unix paths (because `\` is a valid posix char), but in fact it will fix them because it has to be escaped inside the quoted string no matter what it is. 

Nice catch :+1: .

jbalsas: Then we should escape `\` rather than `\\`?

packages/liferay-npm-scripts/src/scripts/webpack.js
```diff
@@ -46,15 +46,19 @@ module.exports = function(...args) {
 	}
 };
 
+function escapeLiteralString(str) {
+	return str.replace(/\\/g, '\\\\');
```
jbalsas: See my comment above... should we replace `\` rather than `\\`?

jbalsas: Ugh, that's what this is doing, right? I always get confused here 😂

izaera: Yes. Backslash escaping is a pain :-S.

Issue #329, opened 2020-01-10
Jest tests are failing when a project imports a local .soy.js file


izaera:
Because we have moved resulting .soy.js files to the build directory, jest tests don't work any more as they are run on the source folder and .js files may import they counterpart .soy.js which, instead of being in the source folder is in the build one.

This is happening since version 17.0.0 when we decided to move all temporary and generated files out of the source folder to avoid deployFast from triggering extra builds.

PR #330, opened 2020-01-10
fix: jest tests are failing when a project imports a local .soy.js file (#329)

izaera:


izaera:
Self-merged after private conversation, as it is urgent.

wincent:
So the old workaround for this kind of thing prior to v17.0.0 involved [this fake transform for soy files](https://github.com/liferay/liferay-npm-tools/blob/944ddde12658e941fc8c38b5af53ab3c761f4d12/packages/liferay-npm-scripts/src/jest/transformSoy.js). It was added in: https://github.com/liferay/liferay-npm-tools/commit/3c5a9c5c217e167394d1771136d8002702919bb4 (see the commit message there for context).

Maybe we don't need most of that anymore, because with this change there will always be an actual JS file on disk that we can resolve and use.

PR #331, opened 2020-01-13
fix: avoid duplicating current project's dependencies (#329)

izaera:


packages/liferay-npm-scripts/src/utils/generateSoyDependencies.js
```diff
@@ -34,7 +37,7 @@ function generateSoyDependencies(dependencies) {
 			return resolvedDependency;
 		})
 		.filter(Boolean)
-		.filter(dependencyPath => dependencyPath !== cwd);
+		.filter(dependencyPath => !dependencyPath.endsWith(projectName));
```
wincent: My only concern here is that there might be some project where the dependency does end with `projectName` but it doesn't correspond exactly to the basename (eg. `projectName` might be `bar`, and `dependencyPath` might be `foo-bar`). So to be fully robust the test should be:

```js
filter(dependencyPath => path.basename(dependencyPath) !== projectName);
```
```

izaera: Good catch

Issue #303, opened 2019-11-25
Set @babel/preset-react as our default babel preset


jbalsas:
We currently have `env` as our default babel preset which forces as to override in any "react" module with:

#### .babelrc

```json
{
	"presets": [
		"@babel/preset-react"
	]
}
```

I currently count  **31 custom .babelrc** in `master` with the react preset, which likely means it's time for us to make this the default.

wincent:
Upstream: https://github.com/brianchandotcom/liferay-portal/pull/83438

PR #335, opened 2020-01-16
feat(scripts)!: make Babel use preset-react by default (#303)

wincent:
Note that this will require some corresponding changes in liferay-portal, namely, there are two projects which use the incremental-dom plugin:

- https://github.com/liferay/liferay-portal/blob/1f84f9713487704f173e6777781a20d7c38477f6/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-builder/.babelrc
- https://github.com/liferay/liferay-portal/blob/1f84f9713487704f173e6777781a20d7c38477f6/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-web/.babelrc

I don't think we can turn off the preset for those, but we can disable the individual plugins that the preset activates.

I believe all the rest should be fine, but will obviously have to test things.

Closes: https://github.com/liferay/liferay-npm-tools/issues/303

### DIFF
--- a/packages/liferay-npm-scripts/src/config/babel.json
+++ b/packages/liferay-npm-scripts/src/config/babel.json
@@ -4,7 +4,7 @@
 			"plugins": ["transform-react-remove-prop-types"]
 		}
 	},
-	"presets": ["@babel/preset-env"],
+	"presets": ["@babel/preset-env", "@babel/preset-react"],
 	"plugins": [
 		"@babel/proposal-class-properties",
 		"@babel/proposal-export-namespace-from",
--- a/packages/liferay-npm-scripts/src/utils/getMergedConfig.js
+++ b/packages/liferay-npm-scripts/src/utils/getMergedConfig.js
@@ -29,6 +29,13 @@ function pluck(config, property) {
 	return config[property];
 }
 
+function isObject(maybeObject) {
+	return (
+		maybeObject &&
+		Object.prototype.toString.call(maybeObject) === '[object Object]'
+	);
+}
+
 /**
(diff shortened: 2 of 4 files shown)


wincent:
Marking this one as a draft until I run a bunch of tests on it in liferay-portal.

wincent:
Hmm. Disabling the plugins may not be as easy as I thought... https://github.com/babel/babel/issues/3016 — I may have to do something a little hackier and custom in [out deepMerge code](https://github.com/liferay/liferay-npm-tools/blob/0e13afcc65799e4032396865ed69e287675ba166/packages/liferay-npm-scripts/src/utils/deepMerge.js#L132). 😢

wincent:
Maybe, just maybe, https://babeljs.io/docs/en/babel-preset-env/#exclude might work.

wincent:
Moving this out of draft status (see "Test plan" in commit message here: https://github.com/liferay/liferay-npm-tools/pull/335/commits/bf2af2e416f756a28cc6afb03cf49c6c6f2e968d).

wincent:
Sadly, after two `ant all` I have realized that it seems I have to release a beta in order to properly test this with a full `ant all` run (because if I clean and then run, then I end up blowing away my local test version with a clean version downloaded from NPM). So I am going to cut a beta and give it another try.

wincent:
`ant all` looks good and manually testing in running liferay-portal instance seems solid. Testing in CI here: https://github.com/wincent/liferay-portal/pull/120

wincent:
I shared this in Slack and received many gifs in reply, which I am taking as a cue to merge. 😀

Issue #339, opened 2020-01-17
Enforce use of .babelrc.js, .prettierrc.js and .eslintrc.js over alternative config file names


wincent:
This is related to #303 ("Set @babel/preset-react as our default babel preset").

It's not enough to just make it the default; we also have to lint to make sure that people don't specify the default preset redundantly (see [#130 in eslint-config-liferay repo](https://github.com/liferay/eslint-config-liferay/issues/130)).

But, to most straightforwardly lint the `.babelrc` files, we have to switch to `.babelrc.js` instead (which will give us formatting for free too).

So, this ticket is about teaching liferay-npm-scripts to enforce that we always use the right config file names.

wincent:
Upstream: https://github.com/brianchandotcom/liferay-portal/pull/83438

PR #340, opened 2020-01-17
feat(scripts)!: enforce use of canonical config file names

wincent:
We want developers to only use `.babelrc.js`, `.eslintrc.js` (and technically, `.prettierrc.js` too, although there should be exactly one of those in liferay-portal) so that we can format and lint those files, to make sure we get the full benefit of rules like this one:

https://github.com/liferay/eslint-config-liferay/blob/master/plugins/eslint-plugin-liferay-portal/docs/rules/no-explicit-extend.md

To that effect, this commit adds an additional "preflight" step to our "check" script, which means that we now have 3 phases:

1. Preflight checks (currently, these just check for disallowed config file names, but this would be a logical spot to put additional checks in the future).
2. Formatting checks (ie. Prettier).
3. Lint checks (ie. ESLint).

I initially considered baking the check for config files into the other steps (ie. the Prettier step could check for bad Prettier config files, and the lint step could check for bad ESLint config files), but that would have involved scanning the repo multiple times and I figure it is better to do one scan (which takes a few seconds) up front. It also ends up being cleaner because we don't have to juggle multiple error conditions and produce clear output for them; eg:

- Config is bad but tool succeeded.
- Config is good but tool failed.
- Config is bad and tool failed.
- Config is good and tool succeeded.

So, we just run three steps (preflight, format, lint) and the logic ends up being very simple.

In any case, we don't have much choice anyway. If we didn't do this, where would we put the enforcement of the use of ".babelrc.js" files? We would need to add a separate step for that anyway. As a bonus, this gives us an obvious extension point for any other non-ESLint checks that we might want to run in the future.

Test plan: Running this in liferay-portal gives output like this in the presence of errors:

    $ liferay-npm-scripts check
    Preflight check failed:
    src/.eslintrc: BAD - use .eslintrc.js instead
    src/.prettierrc: BAD - use .prettierrc.js instead

    Prettier checked 11 files

    1 of 3 jobs failed
    error Command failed with exit code 1.

And in a clean run:

    $ liferay-npm-scripts check
    Prettier checked 11 files

    ✨  Done in 1.70s.

Closes: https://github.com/liferay/liferay-npm-tools/issues/339

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/filterGlobs.js
+++ b/packages/liferay-npm-scripts/src/utils/filterGlobs.js
@@ -19,9 +19,11 @@ function filterGlobs(globs, ...extensions) {
 		}
 	});
 
-	return globs.filter(glob => {
-		return extensions.some(extension => glob.endsWith(extension));
-	});
+	return extensions.length
+		? globs.filter(glob => {
+				return extensions.some(extension => glob.endsWith(extension));
+		  })
+		: globs;
 }
 
 module.exports = filterGlobs;
--- a/packages/liferay-npm-scripts/src/utils/getPaths.js
+++ b/packages/liferay-npm-scripts/src/utils/getPaths.js
@@ -24,7 +24,7 @@ const readIgnoreFile = require('../utils/readIgnoreFile');
 function getPaths(globs, extensions, ignoreFile) {
 	const root = findRoot();
 
-	ignoreFile = path.join(root || '.', ignoreFile);
+	ignoreFile = ignoreFile ? path.join(root || '.', ignoreFile) : '';
 
 	const ignores = fs.existsSync(ignoreFile) ? readIgnoreFile(ignoreFile) : [];
 
--- a/packages/liferay-npm-scripts/src/scripts/check.js
+++ b/packages/liferay-npm-scripts/src/scripts/check.js
@@ -5,11 +5,13 @@
  */
 
 const spawnMultiple = require('../utils/spawnMultiple');
+const preflight = require('./check/preflight');
 const format = require('./format');
 const lint = require('./lint');
 
 function check() {
 	spawnMultiple(
+		() => preflight(),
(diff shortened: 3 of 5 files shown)


packages/liferay-npm-scripts/src/utils/filterGlobs.js
```diff
@@ -19,9 +19,11 @@ function filterGlobs(globs, ...extensions) {
...
 	});
 
-	return globs.filter(glob => {
-		return extensions.some(extension => glob.endsWith(extension));
-	});
+	return extensions.length
+		? globs.filter(glob => {
+				return extensions.some(extension => glob.endsWith(extension));
+		  })
+		: globs;
```
wincent: This tweak allows me to call `getPaths` without any extensions, so that I can grab all the bad config files regardless of extension.

packages/liferay-npm-scripts/src/utils/getPaths.js
```diff
@@ -24,7 +24,7 @@ const readIgnoreFile = require('../utils/readIgnoreFile');
 function getPaths(globs, extensions, ignoreFile) {
 	const root = findRoot();
 
-	ignoreFile = path.join(root || '.', ignoreFile);
+	ignoreFile = ignoreFile ? path.join(root || '.', ignoreFile) : '';
```
wincent: This tweak allows me to not provide an ignore file to `getPaths`. Basically, we'll end up finding *all* the bad configs in the repo, except for those inside `node_modules` (because we always ignore stuff in there).

When `ignoreFile` is empty (`''`), `fs.existsSync()` returns `false`, and we wind up with an empty `ignores` array, which is what we want.

Further down, `filterChangedFiles` will run, which ensures that in CI we don't error out when somebody submits a PR that doesn't even touch one of these bad config files.

izaera:
LGTMYCMIN

PR #341, opened 2020-01-17
chore(scripts)!: update eslint-config-liferay to v16.0.0

wincent:
(Weak) breaking change because it means we'll complain about redundant presets that we didn't complain about in the past.

https://github.com/liferay/eslint-config-liferay/releases/tag/v16.0.0


wincent:
#trivial so going to merge.

PR #345, opened 2020-01-17
chore(scripts): clean up straggling (but inconsequential) .babelrc references

wincent:
These are places where we still talk about ".babelrc" even though we have switched to using ".babelrc.js" files now; none of them should have any meaningful impact on runtime behavior:

- Text in an error message.
- Text in an `it` test description.
- A code comment.

If you `git grep babelrc` you will still find some additional references, but these can stay as they are:

- `runBabel()` still creates temporary .babelrc files, but these always go in a temporary directory (via `withTempFile()`).
- Our storybook wrapper creates .babelrc files too, but always in a temporary directory.
- The bundler preset still has ".babelrc" keys in it, which don't refer to filenames but rather to Babel config in the abstract sense; see: https://github.com/liferay/liferay-js-toolkit/wiki/.npmbundlerrc-file-reference

### DIFF
--- a/packages/liferay-npm-scripts/README.md
+++ b/packages/liferay-npm-scripts/README.md
@@ -117,7 +117,7 @@ If you just set dependencies to be `['my-new-dependency']`, it will override the
 
 ### Other Config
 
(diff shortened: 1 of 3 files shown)


wincent:
#NoLogicChanges, so going to merge.

Issue #336, opened 2020-01-16
Reduce Yarn console warnings


wincent:
Many Yarn operations in liferay-portal trigger a bunch of warnings like these (truncated):

```
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > frontend-taglib-clay > @clayui/icon@3.0.1" has unmet peer dependency "react@^16.8.1".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > frontend-taglib-clay > @clayui/icon@3.0.1" has unmet peer dependency "react-dom@^16.8.1".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > asset-taglib > @clayui/multi-select@3.2.0" has unmet peer dependency "@clayui/css@3.x".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > asset-taglib > @clayui/multi-select@3.2.0" has unmet peer dependency "react-dom@^16.8.1".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > portal-workflow-metrics-web > @clayui/drop-down@3.1.2" has unmet peer dependency "@clayui/css@3.x".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > data-engine-taglib > @clayui/label@3.0.1" has unmet peer dependency "@clayui/css@3.x".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > data-engine-taglib > @clayui/label@3.0.1" has unmet peer dependency "react-dom@^16.8.1".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > data-engine-taglib > @clayui/link@3.0.1" has unmet peer dependency "@clayui/css@3.x".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > data-engine-taglib > @clayui/link@3.0.1" has unmet peer dependency "react-dom@^16.8.1".
warning "workspace-aggregator-2b79b1c4-2309-4d79-8366-92b4c997e755 > data-engine-taglib > @clayui/list@3.0.4" has unmet peer dependency "@clayui/css@3.x".
[...]

PR #346, opened 2020-01-17
refactor(scripts): eliminate many yarn peer dependency warnings (#336)

wincent:
This commit doesn't address the entire problem described in #336 but it is an initial start: namely, it reduces the number of peer dependency warnings that you see when doing a Yarn operation in this repo itself from 93 to ~22 warnings~ 0 warnings.

Original message follows.

---

As you can see in the lockfile, we're not actually changing the version of `@babel/core` that we use here (7.8.3); we're just making explicit the one that we were already using as a result of our transitive dependencies.

In liferay-portal itself, we end up using 7.5.5, so when we update liferay-portal, we'll end up updating that to 7.8.3, which I believe should be a safe change:

    => Found "@babel/core@7.5.5"
      - "workspace-aggregator-7751601e-168c-4244-9341-f549a227cad6" depends on it
      - Hoisted from "_project_#liferay-npm-scripts#babel-jest#@jest#transform#@babel#core"
      - Hoisted from "_project_#liferay-npm-scripts#@storybook#react#@svgr#webpack#@babel#core"
      - Hoisted from "_project_#liferay-npm-scripts#@storybook#react#babel-plugin-react-docgen#react-docgen#@babel#core"
      - Hoisted from "_project_#liferay-npm-scripts#jest#jest-cli#jest-config#@babel#core"
      - Hoisted from "_project_#liferay-npm-scripts#@storybook#react#@svgr#webpack#@svgr#plugin-jsx#@babel#core"

Before (93 warnings):

https://gist.github.com/wincent/18b13750a3db2e2c5d39e8e0a214ed28

After (22 warnings):

https://gist.github.com/wincent/f60645483c0ad0d49549d47796996750

Test plan:

1.  Perform update in liferay-portal:

        yarn add -W --dev ~/code/liferay-npm-tools/packages/liferay-npm-scripts

2.  Verify that duplicate versions go away with a little bit of help:

        npx yarn-deduplicate yarn.lock

Related: https://github.com/liferay/liferay-npm-tools/issues/336


wincent:
Ah, one thing I forgot to mention in the commit message: this change seems like a no-brainer because we very obviously have a clear runtime dependency on `@babel/core`; but the further you go down the list, the more random things get and it becomes less clear that it makes sense to add more explicit dependencies to the package.json just to silence the warnings.

For example, the React-related warnings are there only because of `storybook`, but we are only using Storybook extremely lightly in liferay-portal right now:

```
$ find . -name stories
./modules/dxp/apps/portal-search-tuning/portal-search-tuning-synonyms-web/test/stories
./modules/dxp/apps/portal-search-tuning/portal-search-tuning-rankings-web/test/stories
```

If we add `react` and `react-dom` to the list here then we'll want to do it with a pretty lose version range so as not to cause headaches with duplicate versions later on when integrating (ie. because `frontend-js-react-web` uses exact version numbers).

And the further you get down the list, the more random things get.  I think I'm going to try adding a few more pretty loosely-versioned requirements here and see how far we get, but I don't know if getting to zero warnings in this project (and less so in liferay-portal) is going to happen. 🤔

wincent:
Note there are no longer any warnings in the CI run output:

- https://travis-ci.org/liferay/liferay-npm-tools/jobs/638455720
- https://travis-ci.org/liferay/liferay-npm-tools/jobs/638455720

See under `$ ./.travis/install.sh`:

```
$ ./.travis/install.sh
yarn install v1.15.2
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.11: The platform "linux" is incompatible with this module.
info "fsevents@1.2.11" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
$ ./.travis/script.sh
```

Compared with [an older run](https://travis-ci.org/liferay/liferay-npm-tools/jobs/638403368).

wincent:
Small dep tweaks only, so going to merge this.

PR #347, opened 2020-01-20
feat(scripts): add stylelint

wincent:
~Marking this as a draft until our rule list is defined.~

---

Currently doing nothing because we don't have a set of rules defined yet, but you can see a couple of sample rules commented out that I used for testing.

As noted in previous commits, the API is, unfortunately, async only, which means we need to change a bunch of other places to make them async-capable.

Some other limitations are noted inline, the most obvious of which is that stylelint has an autofix functionality that is very loudly labeled as "experimental":

> Note: It's an experimental feature. It currently does not respect special comments for disabling stylelint within sources (e. g. `/* stylelint-disable */`). Autofixing will be applied regardless of these comments.
>
> If you're using both these special comments and autofixing, please run stylelint twice as a temporary solution. On the first run, some violations could be missed, or some violations might be reported incorrectly.

Source: https://stylelint.io/user-guide/cli#autofixing-errors

Another wrinkle is that stylelint doesn't report whether a particular error is autofixable, so our "fixable" counts are all going to be wrong.

Test plan: Add some errors in JS and JSP files in liferay-portal to make sure those lints still work, then turn on the commented-out stylelint rules. See output like this:

https://gist.github.com/wincent/3d9f2da9a4eed0eedf8ef2ac5d0bc29e

An excerpt:

    modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/js/FragmentsEditor.es.js
      44:1  error  Unexpected 'debugger' statement.  no-debugger

    modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/view_soy.jsp
      56:1  error  Unexpected 'debugger' statement.  no-debugger
      70:1  error  Unexpected 'debugger' statement.  no-debugger

    modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/css/ColumnOverlayGrid.scss
      12:10  error  Unexpected unit "vh" (unit-whitelist)  unit-whitelist
      16:7  error  Unexpected unit "vh" (unit-whitelist)  unit-whitelist

    ...

    ✖ 300 problems (300 errors, 0 warnings)

    2 of 3 jobs failed

### DIFF
--- a/packages/liferay-npm-scripts/src/jsp/lintJSP.js
+++ b/packages/liferay-npm-scripts/src/jsp/lintJSP.js
@@ -72,19 +72,21 @@ function lintJSP(source, onReport, options = {}) {
 				return true;
 			});
 
+			// Ensure trailing newline, matching Prettier's convention.
+			source = output.endsWith('\n') ? output : `${output}\n`;
+
 			if (messages.length) {
 				onReport({
 					errorCount,
 					fixableErrorCount,
 					fixableWarningCount,
 					messages,
-					source: output,
+					source,
 					warningCount
 				});
 			}
 
-			// Ensure trailing newline, matching Prettier's convention.
-			return output.endsWith('\n') ? output : `${output}\n`;
+			return source;
 		}
 	});
 }
--- a/packages/liferay-npm-scripts/src/scripts/lint.js
+++ b/packages/liferay-npm-scripts/src/scripts/lint.js
@@ -282,21 +282,12 @@ async function lintSCSS(source, onReport, options = {}) {
 	const fixableErrorCount = 0;
 	const fixableWarningCount = 0;
 
+	const config = getMergedConfig('stylelint');
+
 	const {output, results} = await stylelint.lint({
 		code: source,
 		codeFilename: filePath,
-		config: {
-			rules: {
-				// TODO: actually decide which rules we want to run; see:
-				// https://github.com/liferay/liferay-npm-tools/issues/344#issuecomment-576196066
-
-				// To test autofix uncomment this:
-				// 'unit-case': 'upper',
-
-				// To test a non-autofixing rule, uncomment this:
-				'unit-whitelist': ['em', 'rem', '%', 's']
-			}
-		},
+		config,
 
 		// Beware: if `fix` is true, stylelint will return the autofixed
 		// source as `output`; when false, it will return an object
--- a/packages/liferay-npm-scripts/src/scripts/lint.js
+++ b/packages/liferay-npm-scripts/src/scripts/lint.js
@@ -7,11 +7,13 @@
 const eslint = require('eslint');
 const fs = require('fs');
 const path = require('path');
+const stylelint = require('stylelint');
 
 const isJSP = require('../jsp/isJSP');
 const lintJSP = require('../jsp/lintJSP');
 const getMergedConfig = require('../utils/getMergedConfig');
 const getPaths = require('../utils/getPaths');
+const hasExtension = require('../utils/hasExtension');
 const log = require('../utils/log');
 const {SpawnError} = require('../utils/spawnSync');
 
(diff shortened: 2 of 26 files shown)


packages/liferay-npm-scripts/src/jsp/lintJSP.js
```diff
@@ -72,19 +72,21 @@ function lintJSP(source, onReport, options = {}) {
 				return true;
 			});
 
+			// Ensure trailing newline, matching Prettier's convention.
+			source = output.endsWith('\n') ? output : `${output}\n`;
+
```
wincent: Moved this up here for consistency with the equivalent code in `lintSCSS`; it seems that the `source` we report via the `onReport` callback should be the same as the string we return from this function.

packages/liferay-npm-scripts/src/scripts/lint.js
```diff
@@ -238,26 +263,131 @@ function formatter(results) {
...
+	// totally reliable way to guess, so these counts will always be
+	// (possibly innaccurately) zero.
+	const fixableErrorCount = 0;
+	const fixableWarningCount = 0;
+
+	const {output, results} = await stylelint.lint({
+		code: source,
+		codeFilename: filePath,
+		config: {
+			rules: {
```
jbalsas: Isn't there a way to provide the rules via config as we do for the other tools?

jbalsas:
LGTM! Only thing I think might be missing was how to configure/override `stylelint` config. What do we have in mind for that?

packages/liferay-npm-scripts/src/scripts/lint.js
```diff
@@ -85,25 +101,34 @@ function lint(options = {}) {
...
+						});
+					},
+					{filePath, fix, quiet}
+				);
+
+				if (fix && updated !== contents) {
+					fs.writeFileSync(filePath, updated);
+				}
+			} catch (error) {
+				log(error);
```
wincent: I don't expect to see errors here (old code path didn't both `catch`-ing them, and I didn't see any in testing) but doing this out of an excess of caution seeing as I don't know how trustworthy the stylelint codebase is yet. At least in initial testing stylelint seems to degrade pretty gracefully; for example, a syntax error gets reported like any other lint error:

```
modules/apps/layout/layout-content-page-editor-web/src/main/resources/META-INF/resources/css/main.scss
  98:6  error  Unexpected } (CssSyntaxError)  CssSyntaxError
```

packages/liferay-npm-scripts/src/scripts/lint.js
wincent: There [are `.stylelintrc` (etc) files](https://stylelint.io/user-guide/configuration), although I have not tested to see whether the Node.js API reads them automatically. In any case, I was thinking of sticking the defaults [in here with all the others](https://github.com/liferay/liferay-npm-tools/tree/master/packages/liferay-npm-scripts/src/config).

packages/liferay-npm-scripts/src/scripts/lint.js
jbalsas: Interesting... is that a `CSS Syntax Error` or an `SCSS Syntax Error`? Are we compiling then parsing?

wincent: I inserted a bunch of extra `}}}}` in an ".scss" file.

We're only linting ".scss" files, and we're passing `syntax: 'scss'` to stylelint, so it is doing the parsing.

wincent:
I pushed a couple of extra commits showing how we can read default config from a bundled file and merge it with whatever is on disk (probably nothing). A sample run with these rules:

```json
{
	"rules": {
		"color-hex-case": "lower",
		"color-hex-length": "short",
		"color-no-invalid-hex": true,
		"no-duplicate-at-import-rules": true,
		"number-leading-zero": "never",
		"number-no-trailing-zeros": true,
		"unit-case": "lower"
	}
}

```

Reports these (autofixable) "errors":

https://gist.github.com/wincent/74d6192d4b68c21841fd9cb4b6b07ca9

And yields this diff:

https://gist.github.com/wincent/1a03dbd38e9502fd81816c11516658a3

wincent:
Took this out of draft status because I've done some reasonable-looking test runs in liferay-portal:

- [Initial run](https://gist.github.com/wincent/3159cd312bc330767a8cd5804e6d55ac) (2954 problems).
- [Autofix run](https://gist.github.com/wincent/a4968749a23be475f44b1f7498cf6ac3) (1261 problems — don't know why but it seems it underreports errors when fixing... 🤔).
  - [Autofix diff](https://gist.github.com/wincent/c8780bbd49b6b9b821c7135ceeb45adc) (regrettably quite small; the whole diff is 553 lines).
- [Re-check](https://gist.github.com/wincent/0c7bf7d3a7dbf575a23ea25d7b7fca78) (2890 problems).
- [After adding at-rule config, turning off declaration no-important](https://gist.github.com/wincent/7a50cfa3d736d4158bbd6e49a714aedf) (2008 problems).
- [After turning off string-quotes](https://gist.github.com/wincent/2ca8da808f2ce03717b85c348ef58451) (1198 problems).
- [After turning off vendor prefix rules](https://gist.github.com/wincent/526593a41b913dfc81564de5da9f8014) (1048 problems).
- [After fixing number-leading-zero with a regex](https://gist.github.com/wincent/972076dac693f1afcc55e474efe057ec) (222 problems)

And after that, inverting the sense of the function-url-quotes rule gets rid of another 98 errors, so I think the remaining ones can be dealt with when we merge this into liferay-portal.

wincent:
Going to do a quick rebase of this to resolve the conflict in `fix.js` and will merge it once I see CI green.

PR #348, opened 2020-01-21
feat(scripts)!: check desired filename format

wincent:
Draft only because this is just a proof-of-concept at this point because we haven't actually decided on the rules that we want to enforce.

---

Note that I am just doing this for SCSS files at this point because we might even want to pursue the alternative of baking this directly into our ESLint rules for our JS files.

FWIW, a quick survey of existing names in liferay-portal shows:

- 68 files using uppercase (ie. CamelCase) names: https://gist.github.com/wincent/6d62f155c01dee91c9dd5f0f6a533d8c
- 104 files using kebab-case names: https://gist.github.com/wincent/c7d39715cba6f794f558b71538bdc12a
- 758 files using neither of the above (generally, that means 0 or more underscores): https://gist.github.com/wincent/df6e2f276d91bb34a89cb17c7cd8ee99

Test plan: run in liferay-portal; each `getPaths` call takes about 4 seconds on my machine even though that means scanning the entire repo -- we could memoize that or make it "smarter" in CI (ie. by using the changed paths detection to short-circuit traversal instead of doing it at the end as a filtering step).

Related: https://github.com/liferay/liferay-frontend-guidelines/issues/78

### DIFF
--- a/packages/liferay-npm-scripts/src/scripts/check/preflight.js
+++ b/packages/liferay-npm-scripts/src/scripts/check/preflight.js
@@ -6,6 +6,7 @@
 
 const path = require('path');
 
+const getMergedConfig = require('../../utils/getMergedConfig');
 const getPaths = require('../../utils/getPaths');
 const log = require('../../utils/log');
 const {SpawnError} = require('../../utils/spawnSync');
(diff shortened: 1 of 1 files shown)


wincent:
In the absence of consensus about desired file-naming pattern, going to close this. Can resurrect in the future if needed.

PR #350, opened 2020-01-22
chore(scripts): update eslint-config-liferay to v17.0.0

wincent:
Provides us with a new "no-loader-import-specifier" rule which will enforce the following:

```js
// Incorrect code:
import useless from './styles.scss';

// Correct code:
import './styles.scss';
```

See: https://github.com/liferay/eslint-config-liferay/releases/tag/v17.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"devDependencies": {
 		"eslint": "6.8.0",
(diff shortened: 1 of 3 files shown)


wincent:
Just a routine update so I'll merge this as soon as I see CI green.

PR #356, opened 2020-01-24
chore: remove unnecessary line from .prettierignore

wincent:
Because Prettier ignores node_modules by default:

> Prettier CLI will ignore files located in `node_modules` directory. To opt-out from this behavior use `--with-node-modules` flag.

https://prettier.io/docs/en/cli.html


wincent:
Trivs.

PR #358, opened 2020-01-30
fix(scripts): tighten up Jest module name mapper generation

wincent:
Latent bug in here that I don't think has bitten us in practice, but while testing changes for [LPS-98114](https://issues.liferay.com/browse/LPS-98114), I noticed that I could change imports in the following ways without triggering expected errors:

    import {foo} from 'some-module';

    // Add a prefix:
    import {bar} from 'sssssssome-module';

    // Add a suffix:
    import {baz} from 'some-moduleeeee';

We need to use stricter matching. The Jest documentation itself points out this pitfall:

> Note: If you provide module name without boundaries `^$` it may cause hard to spot errors. E.g. `relay` will replace all modules which contain `relay` as a substring in its name: `relay`, `react-relay` and `graphql-relay` will all be pointed to your stub.

See: https://jestjs.io/docs/en/configuration#modulenamemapper-objectstring-string

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/getJestModuleNameMapper.js
+++ b/packages/liferay-npm-scripts/src/utils/getJestModuleNameMapper.js
@@ -91,7 +91,9 @@ function getJestModuleNameMapper() {
 									`^${basename}/(.*)`
 								] = `<rootDir>${relative}/$1`;
 							} else {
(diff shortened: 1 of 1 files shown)


wincent:
Very trivial, so going to merge this.

PR #360, opened 2020-01-30
fix(scripts): make soy resolver overrides work with ".soy" as well

wincent:
We have a mix of these two import styles in liferay-portal:

    import templates from './Thing.soy';
    import templates from './Other.soy.js';

- ".soy": about 93 hits (via `git grep "import.+\.soy'" -- '*.js'`)
- ".soy.js": about 65 hits (via `git grep "import.+\.soy\.js'" -- '*.js'`)

As such, the resolver hack that we had in place for ".soy.js" files was brittle.

In this commit, I make it work for both. If the requested file ends in either extension and passes our `basedir` (ie. own project) check, then we rewrite it, and normalize the extension to always be ".soy.js".

On the source side, I am preparing a commit to liferay-portal that makes us use ".soy" everywhere; in general we don't want an explicit ".js" extension in any `import` statement, because it is inconsistent.

I'll follow that up with a lint to ensure that we're don't let any of these creep back in.

Note also that I had to set up `transformIgnorePatterns` because once we have rewritten own-project references, we want to make sure that we require the files that we actually built (with `buildSoy`) rather than falling through to our `transformSoy` stub).

### DIFF
--- a/packages/liferay-npm-scripts/src/config/jest.config.js
+++ b/packages/liferay-npm-scripts/src/config/jest.config.js
@@ -23,5 +23,6 @@ module.exports = {
 		'\\.soy$': path.join(__dirname, '..', 'jest', 'transformSoy.js'),
 		'.+': path.join(__dirname, '..', 'jest', 'transformBabel.js')
 		/* eslint-enable sort-keys */
-	}
+	},
+	transformIgnorePatterns: ['/node_modules/', '<rootDir>/.*\\.soy$']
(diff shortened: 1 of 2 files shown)


wincent:
As noted in https://github.com/liferay/liferay-npm-tools/pull/361, testing at https://github.com/wincent/liferay-portal/pull/127 went well, so going to merge this and cut a release shortly.

PR #361, opened 2020-01-30
chore: prepare liferay-npm-scripts 21.0.1-beta.1 release

wincent:
Just a draft for visibility; I need to make a test release so that I can ask John Co to test the biggish corresponding change in liferay-portal.

PR #363, opened 2020-01-31
fix(scripts): treat webpack.config.js as a config file

wincent:
Originally included in [the beta release](https://github.com/liferay/liferay-npm-tools/pull/361) I did yesterday, but forget to sent it to master as well, so here it goes.

---

So that we don't have to write custom ESLint exclusions just so that we
can use symbols like `require` in the config files. Will silence output
like:

    modules/apps/analytics/analytics-client-js/webpack.config.js
      15:17  error  'require' is not defined.  no-undef
      16:24  error  '__dirname' is not defined.  no-undef
      19:1  error  'module' is not defined.  no-undef

### DIFF
--- a/packages/liferay-npm-scripts/src/config/eslint.config.js
+++ b/packages/liferay-npm-scripts/src/config/eslint.config.js
@@ -9,7 +9,8 @@ const CONFIG_FILES = [
 	'**/.eslintrc.js',
 	'**/.prettierrc.js',
 	'**/.stylelintrc.js',
-	'**/npmscripts.config.js'
+	'**/npmscripts.config.js',
+	'**/webpack.config.js'
 ];
 
 module.exports = {


wincent:
Trivial and previously tested, so going to merge this.

jbalsas:
Nice! This was starting to bother me a little 😂

jbalsas:
Actually, what I think was bothering me is how we need to configure `.eslintrc` in places where we use node.js features or tests... there don't seem to be too many now, so maybe it's just a bad memory :)

wincent:
Yeah, they should be dwindling away at this point. We have ESLint [configured](https://github.com/liferay/liferay-npm-tools/blob/649f5d0b9d3af3dd003670d4b2e8411cb61a5dca/packages/liferay-npm-scripts/src/config/eslint.config.js#L30-L44) to assume `env: 'node'` in `**/test/**/*.js` files. And about the only other place where we need Node.js stuff should be in these config files...

Maybe it is time to do another sweep to remove useless cruft from `.eslintrc.js` config files, of which there are still about 32 such files in liferay-portal...

wincent:
> Maybe it is time to do another sweep to remove useless cruft from .eslintrc.js config files, of which there are still about 32 such files in liferay-portal...
 
Will take care of this in: https://issues.liferay.com/browse/LPS-107950

wincent:
Ok, after auditing liferay-portal, seems we can:

- Totally delete 8 of the 32 `.eslintrc.js` files; and:
- Trim stuff out of 2 others.

Provided that we expand the "config detection" code seen in this PR to also recognize `gulpfile.js` and `webpack.config.{env}.js`. I'll prepare something for that today and ship it on Monday (because I already have on liferay-npm-scripts PR in flight in Brian's fork and I don't want to produce a conflict with myself).

wincent:
Follow-up for Monday is here: https://github.com/liferay/liferay-npm-tools/pull/364

PR #367, opened 2020-02-06
feat(scripts): expose "prettier" subcommand

wincent:
As explained in the README, the intent here is to facilitate editor integration with our customized spin on Prettier. The idea is that instead of using the `prettier` executable, you can use `liferay-npm-scripts prettier` as a more-or-less drop-in replacement.

Next steps will involve showing how to actually use this in some editors, like Vim and VSCode.

### DIFF
--- a/packages/liferay-npm-scripts/contrib/prettier/prettier.js
+++ b/packages/liferay-npm-scripts/contrib/prettier/prettier.js
@@ -984,7 +984,7 @@ module.exports = {
 
 /**
  * Returns the root directory if `filepath` is inside a liferay-portal
- * checkout, otherwise return null.
+ * checkout, otherwise returns null.
  */
 function getPortalRoot(filepath) {
 	// Walk up until we find portal-web (in the root).
--- a/packages/liferay-npm-scripts/contrib/prettier/prettier.js
+++ b/packages/liferay-npm-scripts/contrib/prettier/prettier.js
@@ -1,0 +1,1080 @@
+/**
+ * © 2020 Liferay, Inc. <https://liferay.com>
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+/**
+ * Sample script that you can drop into your path to expose the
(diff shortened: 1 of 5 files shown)


wincent:
Meant to mark this as a draft PR, because I wanted to add another commit (or maybe two) on top to refactor and remove some duplication, but in any case this has passed my basic manual QA at this point. I want to do a bit more testing from inside Vim (and then VSCode) before actually merging this.

wincent:
@jbalsas: This is the stand-in for the `prettier` executable that I was telling you about.

cc @julien in case you want to take a look.

wincent:
As a test run, [this gist](https://gist.github.com/wincent/bddd2270f8fe1ed5cfab7b6d81521eec) shows the result of running `packages/liferay-npm-scripts/bin/liferay-npm-scripts.js prettier '**/*.js' --write` against this repo.

wincent:
I think this is good but will re-test it tomorrow with fresh eyes and if all is good we can cut a release.

packages/liferay-npm-scripts/contrib/prettier/prettier.js
```diff
@@ -0,0 +1,1080 @@
...
+			return prettier.resolveConfig(filepath, options);
+		} else {
+			return Promise.resolve(null);
+		}
+	},
+
+	version: '1.19.1'
+};
+
+const PORTAL_WEB = 'portal-web';
```
wincent: I'll inline this. It's only here in a constant because during an earlier iteration of this code I was using it in two places.

packages/liferay-npm-scripts/contrib/prettier/prettier.js
```diff
@@ -0,0 +1,1080 @@
...
+
+/**
+ * Returns the root directory if `filepath` is inside a liferay-portal
+ * checkout, otherwise return null.
```
wincent: `s/return/returns/`

PR #368, opened 2020-02-10
feat(preset): set up style-loader

wincent:
People are [using this in liferay-portal](https://github.com/brianchandotcom/liferay-portal/commit/dfb71a9dce4924f11912396f9cc1d4849874a8bd
) but we don't want people sneaking in `devDependencies` by declaring them in `dependencies`, so let's make it "official".

About the version number: that liferay-portal commit was using quite a stale version of the loaders, presumably because there was a long delay between it being written and finally being merged. This is another reason why we'd prefer to control the version numbers centrally (via this preset). Note that I'm using 2.17.0 (and not the latest, 2.17.1) just to keep this change small. There was only one small change in 2.17.1, and it doesn't seem relevant to liferay-portal (it was done for the benefit of the liferay-js-themes-toolkit):

https://github.com/liferay/liferay-js-toolkit/commit/a09456d598dac80ef0c9152b04b16c99e3425d69

### DIFF
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -269,6 +269,10 @@
 	"output": "build/node/packageRunBuild/resources",
 	"rules": [
 		{
+			"test": "\\.css$",
+			"use": ["style-loader"]
+		},
+		{
 			"test": "\\.scss$",
 			"exclude": "node_modules",
 			"use": [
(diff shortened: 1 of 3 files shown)


wincent:
Seems we can avoid doing this (and opening the floodgates to potential abuse of the feature) if we cut a new release of the bundler... this fix is required: https://github.com/liferay/liferay-js-toolkit/issues/459

Closing this one in the meantime.

Issue #369, opened 2020-02-10
Add "process" global to ESLint default


wincent:
[Originally posted in the eslint-config-liferay repo](https://github.com/liferay/eslint-config-liferay/issues/118), but it actually belongs here (all of [our default globals are declared here](https://github.com/liferay/liferay-npm-tools/blob/290a1784c368d6882b0ea9e488f005f4bd470046/packages/liferay-npm-scripts/src/config/eslint.config.js#L24-L31), and we have no globals declared in the eslint-config-liferay repo.)

Labeling this as a "feature" because it will enable us to streamline the config over in the [liferay-portal](https://github.com/liferay/liferay-portal) repo.

---

In the old days it wasn't actually safe to assume `process` was defined, because not all code passed through the bundler (which is the one [that sets up Babel](https://github.com/liferay/liferay-js-toolkit/wiki/How-to-use-liferay-npm-bundler#an-example) to use the [babel-plugin-transform-node-env-inline package](https://www.npmjs.com/package/babel-plugin-transform-node-env-inline)). But right now, pretty much everything *does* go through it, and we could afford to add `process: true` to our `globals` set-up in one place instead of sprinkling it around in various `.eslintrc.js` files (as seen in [this PR](https://github.com/brianchandotcom/liferay-portal/pull/80236), and others).

The only exceptions I am aware of right now are our frontend-js-web Webpack bundle, which operates in `mode: 'production'` during the build and inspecting the build output I was able to confirm that `if (process.env.NODE_ENV === '...')` checks are getting correctly compiled away. And the other open question would be what is happening on the backport targets (7.2.x, 7.1.x etc); we would want to confirm that the assumptions hold there too before rolling this out everywhere.

PR #370, opened 2020-02-10
feat(scripts): add "process" to the default list of globals

wincent:
As explained in the related issue:

In the old days it wasn't actually safe to assume `process` was defined, because not all code passed through the bundler, which is the one that [sets up Babel](https://github.com/liferay/liferay-js-toolkit/wiki/How-to-use-liferay-npm-bundler#an-example) to use the [babel-plugin-transform-node-env-inline package](https://www.npmjs.com/package/babel-plugin-transform-node-env-inline).

But right now, pretty much everything *does* go through it, and we can afford to add `process: true` to our `globals` set-up in one place instead of sprinkling it around in various `.eslintrc.js` files.

The only exceptions I am aware of right now are our frontend-js-web Webpack bundle, which operates in `mode: 'production'` during the build and inspecting the build output I was able to confirm that `if (process.env.NODE_ENV === '...')` checks are getting correctly compiled away.

Marking this one as a "feature" because it will streamline the task of configuration in liferay-portal.

Closes: https://github.com/liferay/liferay-npm-tools/issues/369

### DIFF
--- a/packages/liferay-npm-scripts/src/config/eslint.config.js
+++ b/packages/liferay-npm-scripts/src/config/eslint.config.js
@@ -25,6 +25,7 @@ module.exports = {
 		AUI: true,
 		CKEDITOR: true,
 		Liferay: true,
+		process: true,
 		submitForm: true,
 		svg4everybody: true,
 		themeDisplay: true


wincent:
Chatted with @julien about this; he agrees with my assessment that this is relatively safe, so I'm going to merge it.

PR #374, opened 2020-02-11
chore(preset): update to bundler v2.18.0

wincent:

PR #375, opened 2020-02-11
chore(scripts): update to bundler v2.18.0

wincent:
As predicted [here](https://github.com/liferay/liferay-npm-tools/pull/374#issuecomment-584733931), this removes the temporary duplication in the lock file introduced by that pull (necessary because we have a two-step release process for updating the bundler).

All of this because we want [this fix](https://github.com/liferay/liferay-js-toolkit/commit/c0632cd851838f2e98827b4fd931a565197e2d8d).


wincent:
As this is just a dep change, merging this.

PR #377, opened 2020-02-12
feat(scripts): catch banned dependencies in package.json file

wincent:
Please see the individual commit messages for details, but the gist of it is:

- We check against bad dependencies getting added to "package.json" files.
- We format the JSON with Prettier.

Beware: we can't merge this until the bundler issue [described over here](https://github.com/izaera/liferay-portal/pull/125) is fixed, because only then will we be able to remove the bad dependencies from the dynamic-data-mapping-form-field-type project.

### DIFF
--- a/packages/liferay-npm-scripts/src/presets/standard/index.js
+++ b/packages/liferay-npm-scripts/src/presets/standard/index.js
@@ -21,6 +21,9 @@ module.exports = {
 	},
 	check: [...JS_GLOBS, ...JSP_GLOBS, ...SCSS_GLOBS],
 	fix: [...JS_GLOBS, ...JSP_GLOBS, ...SCSS_GLOBS],
+	rules: {
+		'blacklisted-dependency-patterns': ['^liferay-npm-bundler-loader-.+']
+	},
 	storybook: {
 		languagePaths: ['src/main/resources/content/Language.properties'],
 		port: '9000',
--- a/packages/liferay-npm-scripts/src/presets/standard/index.js
+++ b/packages/liferay-npm-scripts/src/presets/standard/index.js
@@ -9,8 +9,16 @@ const liferay = require('./dependencies/liferay');
 const metal = require('./dependencies/metal');
 
 const JS_GLOBS = ['{src,test}/**/*.js'];
+const JSON_GLOBS = ['/*.json'];
 const JSP_GLOBS = ['src/**/*.{jsp,jspf}'];
 const SCSS_GLOBS = ['{src,test}/**/*.scss'];
+
+const CHECK_AND_FIX_GLOBS = [
+	...JS_GLOBS,
+	...JSON_GLOBS,
+	...JSP_GLOBS,
+	...SCSS_GLOBS
+];
 
 module.exports = {
 	build: {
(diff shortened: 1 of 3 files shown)


packages/liferay-npm-scripts/src/presets/standard/index.js
```diff
@@ -9,18 +9,29 @@ const liferay = require('./dependencies/liferay');
 const metal = require('./dependencies/metal');
 
 const JS_GLOBS = ['{src,test}/**/*.js'];
+const JSON_GLOBS = ['/*.json'];
```
jbalsas: I think this will fight Hugo's SF and if I remember correctly, we agreed he'll be the one formatting JSON, did we not?

jbalsas: We can maybe bring it up with Brian if we feel we should take over... or make sure we don't get in a json-formatting-loop-fight

wincent: I believe I confirmed that it doesn't fight it as [described here](https://github.com/liferay/liferay-npm-tools/pull/377/commits/de5d8de4d528f6059687ac11c4cfdacb88c175f3). Sample "ci:test:sf" results run [here](https://github.com/wincent/liferay-portal/pull/131).

I can drop this commit easily enough, but I think the source formatter is somewhat lax about package.json files because they are often machine-edited (ie. by doing `yarn add`, `yarn remove` etc).

wincent: There are still a bunch of JSON (and `package.json`) checks which will run in non-conflicting ways; see [classes here with "JSON" in their names](https://github.com/liferay/liferay-portal/tree/master/modules/util/source-formatter/src/main/java/com/liferay/source/formatter/checks) (`JSONWhitespaceCheck.java`, `JSONPropertyOrderCheck.java` etc).

jbalsas: let's go for it, then! 👍 

PS: Apologies for not reading the thorough commit description... I was skimming through PRs and emails to start the day and didn't stop to review thoroughly 😢

wincent: Another option to consider is that we could "abuse" [the `JSONDeprecatedPackagesCheck` class](https://github.com/liferay/liferay-portal/blob/3f4502ca3f108494b444c1d82598dbe5b8e34bda/modules/util/source-formatter/src/main/java/com/liferay/source/formatter/checks/JSONDeprecatedPackagesCheck.java#L61-L63) to address the "banned dependencies" problem that this PR is aimed at. The only things I don't like about that though is:

- You can't specify packages using a pattern; you have to list out literal names.
- The message the user sees will be "Do not use deprecated package XYZ", which is not strictly accurate: in this PR, the message is "contains blacklisted dependency: XYZ".
- It is ceding a bit of control.

In any case, I have no hurry with this PR because we can't use it until the bundler problem I mentioned in the PR description is fixed.

jbalsas: We can turn the tide around then and bring the logic in `JSONDeprecatedPackagesCheck` here in a way that can be configured so we can delete those and rely on a simple configuration setting in DXP

wincent: seinfeld-i-like-the-way-you-think

wincent:
Ok, new bundler packages are out, so we can prep a release of the scripts. Going to merge this.

PR #378, opened 2020-02-13
chore(preset): update bundler dependencies to v2.18.1

wincent:
Just a routine update, like https://github.com/liferay/liferay-npm-tools/pull/374 — as noted there, the duplication in the lockfile is expected and temporary.


wincent:
No need to wait for full CI on this one; the real integration test comes only once we update liferay-npm-scripts, so going to merge this one.

Issue #380, opened 2020-02-13
Enforce trailing commas


wincent:
Originally requested [here](https://github.com/liferay/liferay-frontend-guidelines/issues/120) and green-lighted [here](https://github.com/liferay/liferay-frontend-guidelines/issues/120#issuecomment-585834606).

wincent:
@jbalsas: Did you ever end up asking brianchandotcom about this?

jbalsas:
As it turns out, I never got around to bring this up... as per our offline conversation, I think we might not see an immediate return here, specially since we wouldn't be able to format JS in JSP this way, so maybe it's best to document this limitation and park this for now?

Full disclosure, this particular feature is one of the ones I struggle with the most... but I wouldn't block it if the majority felt positively about it.

p2kmgcl:
I don't like trailing commas :laughing:, I thought it would be useful to have them in portal. If no one else wants to have them we can ignore this.

wincent:
I feel a bit stupid.

[This MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) claims that we *can* use trailing commas in IE9 and above, so I was wrong about this not being safe to put in our JSPs. This page from incredibly trustworthy/legendary JS educator Axel Rauschmayer also corroborates that only IE 8 doesn't support it...

Evidently, I was thinking of [this problem](https://github.com/brianchandotcom/liferay-portal/pull/79731#issuecomment-543540235) in JSP, which is precisely the opposite of what we're talking about here — that the *lack* of trailing commas in JSP was actually *causing* us problems when it came to conditional execution vs formatting...  🤦‍♂️

PR #381, opened 2020-02-13
feat(scripts)!: enforce trailing commas in JS

wincent:
This wasn't previously safe to do because IE will choke on a trailing comma, but now that basically all JS is passing through Babel and getting transformed away, we can go ahead.

The exceptions to this are theme JS, of which there is almost none, and JS in JSP, which is not affected (because we adjust Prettier to make sure that it never adds commas in that context).

Test plan: In a test run over liferay-portal, produces [this 77k-line diff](https://gist.github.com/wincent/2643a647345a98ffb4df9b559aba3688).

Contrast that with this, where we skip the special casing of JSP, and it would produce [this 88k-line diff](https://gist.github.com/wincent/223f31e62f74cf542bd8f1b9df087826), which would evidently cause problems in IE.

Closes: https://github.com/liferay/liferay-npm-tools/issues/380

Related: https://github.com/liferay/liferay-frontend-guidelines/issues/120

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,19 +9,19 @@ module.exports = {
 	env: {
 		browser: true,
 		jest: true,
-		node: true
+		node: true,
 	},
 	extends: 'liferay',
 	parserOptions: {
-		ecmaVersion: 2018
+		ecmaVersion: 2018,
 	},
 	rules: {
 		'no-for-of-loops/no-for-of-loops': 'off',
 		'notice/notice': [
 			'error',
 			{
-				templateFile: path.join(__dirname, 'copyright.js')
-			}
-		]
-	}
+				templateFile: path.join(__dirname, 'copyright.js'),
+			},
+		],
+	},
 };
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
 	"bracketSpacing": false,
 	"singleQuote": true,
 	"tabWidth": 4,
+	"trailingComma": "es5",
 	"useTabs": true
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,0 @@
-{
-	"bracketSpacing": false,
-	"singleQuote": true,
-	"tabWidth": 4,
-	"trailingComma": "es5",
-	"useTabs": true
-}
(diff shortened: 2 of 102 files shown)


wincent:
I foresee a merge conflict or two in the future...

vision

wincent:
@jbalsas: Do you want to practice your famous diplomacy skills before we try this one out? Here is a sample PR demonstrating the result in liferay-portal in case you wanted to show it to Brian: https://github.com/wincent/liferay-portal/pull/136

jbalsas:
Ok, I'll run it by Brian later today!

wincent:
See [my comment here](https://github.com/liferay/liferay-npm-tools/issues/380#issuecomment-590944485) — we actually can do this in JSP as well as JS, which means we can be consistent.

wincent:
Rebased/updated: now behaves uniformly in both JS and JSP files.

wincent:
Updated https://github.com/wincent/liferay-portal/pull/136 too, showing what the final result of applying this would look like (note: due to GitHub service degradation, the `git push` worked, but the updated commits aren't showing up in the UI yet).

wincent:
Parked this [over here](https://github.com/brianchandotcom/liferay-portal/pull/85253) so that we can run it past Brian.

wincent:
Brian [says this it's fine](https://github.com/brianchandotcom/liferay-portal/pull/85253#issuecomment-591097381). I updated [my PR](https://github.com/wincent/liferay-portal/pull/136) with [this change](https://github.com/wincent/liferay-portal/pull/136/commits/f8469d54c52e4cece94ca8866f8997fbf463378d) that turns off the [conflicting Java source-formatting](https://github.com/brianchandotcom/liferay-portal/pull/85253#issuecomment-590984955) to show that we can pass "ci:test:sf" with this change ([see the results](https://github.com/wincent/liferay-portal/pull/136#issuecomment-591360106), I believe failures are in untouched files).

Note, though, that it is probably *not* the actual fix that we'll need to apply (because it turns off *all* comma-deletion unconditionally in JSP; it's not scoped just to JS script blocks), but it at least shows the ball-park in which we'll need to make the tweak.

wincent:
I think [this is a better fix](https://github.com/wincent/liferay-portal/pull/150/commits/580927f57b2a2fda5e1c663e60a20394f9c03d0d), so I am going to merge this and cut a release so that I can send all this over to Hugo for him to take a look.

PR #385, opened 2020-02-17
chore: update dependencies

wincent:

PR #386, opened 2020-02-17
refactor(reporter): remove explicit devDependencies

wincent:
We shouldn't have explicit devDependencies in our monorepos; they should all be shared and declared at the top level.

Jest already is, so everything continues to work exactly as before (and note: no changes to the lock file).


wincent:
Splitting this off from https://github.com/liferay/liferay-npm-tools/pull/385 because it is lower risk, and if I want to cut a beta to properly test that one I can do that without also having to cut a beta of this dependency.

wincent:
This one is trivial, so merging.

PR #387, opened 2020-02-17
revert: undo reporter changes related to IFI-780

wincent:
Given [John's comment on IFI-780](https://issues.liferay.com/browse/IFI-780):

> This solution does not resolve the root issue.

So, let's revert b50a923 and the follow-ups (418f691 and d52cc87). If we want to take another stab at this in the future we should start again. At the moment, our last non-beta release was 1.0.1 and we'd like to make another (see https://github.com/liferay/liferay-npm-tools/pull/386), but the latest on "master" is 1.1.0-beta.1 and we can't release on top of that. So, we roll back the version change as well (a268cb4) to clear the way for a subsequent release. We'll bump the version number all the way to v1.2.0 to avoid the confusion that may arise from having a "latest" version that is a lower number than a prior release from many months ago.

### DIFF
--- a/packages/liferay-jest-junit-reporter/package.json
+++ b/packages/liferay-jest-junit-reporter/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "liferay-jest-junit-reporter",
-	"version": "1.1.0-beta.1",
+	"version": "1.0.1",
 	"description": "A JUnit reporter for Jest and Liferay CI",
 	"main": "src/index.js",
 	"scripts": {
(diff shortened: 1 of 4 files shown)


wincent:
Semantic PR bot is complaining about 9095b42, because that reverts an ancient commit that didn't follow the Conventional Commits pattern; FWIW, I think it's better to ignore the "error" and refrain from mutating the old commit message in this case.

wincent:
For visibility, after this revert, diff between v1.0.1 and `HEAD` of `master` will be this: https://gist.github.com/wincent/dca2732a5867fed3702c8ddbff9d5db8

Issue #389, opened 2020-02-19
Sort SCSS imports


wincent:
Previously requested [here](https://github.com/brianchandotcom/liferay-portal/pull/81633#issuecomment-561865490). We were tracking SCSS formatting in a general fashion [here](https://github.com/liferay/liferay-npm-tools/issues/344), but that is closed now and it's time to create a new issue for this here as it just came up [again](https://github.com/brianchandotcom/liferay-portal/pull/84932#issuecomment-588351493).

As is elsewhere the convention in liferay-portal, the idea is to treat empty lines as boundaries across which no sorting-related movement may occur, but otherwise, we will sort lexicographically. This might break stuff until we flush out the bugs (and add empty lines where needed), but once the dust settles it should be safe.

PR #390, opened 2020-02-20
feat(script)!: sort imports in SCSS files

wincent:
We sort alphabetically, with blank lines used to divide imports into groups. We only sort within groups, never across groups, which means that you can force things to be imported in a certain order if necessary.

Example output on liferay-portal:

https://gist.github.com/wincent/d5e48a398930b327c9cc29291ee35dc3

Example autofix:

https://gist.github.com/wincent/c9dd73cda61e2c3519a280301ba2458a

Note that in there I inserted a blank-line after an `@import "variables"` in page_editor just to show how this can be used to prevent an autosort from being applied; if you look in another file in page_editor in that diff, you'll see what happens if you don't do that.

Closes: https://github.com/liferay/liferay-npm-tools/issues/389

### DIFF
--- a/packages/liferay-npm-scripts/README.md
+++ b/packages/liferay-npm-scripts/README.md
@@ -36,7 +36,7 @@ Do you need to use `liferay-npm-bridge-generator`? Just add a `.npmbridgerc` fil
 liferay-npm-scripts check
 ```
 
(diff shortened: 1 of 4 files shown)


wincent:
Given the test result, I think this one is safe to merge.

Issue #391, opened 2020-02-20
Disallow explicit extension in SCSS imports


wincent:
While preparing https://github.com/brianchandotcom/liferay-portal/pull/85026 I just noticed a superfluous ".scss" extension in an import statement. We should lint and autofix that.

wincent:
There is a library that we could use for this which has [an `at-import-partial-extension` rule](https://github.com/kristerkari/stylelint-scss/blob/master/src/rules/at-import-partial-extension/README.md), but we should evaluate that dependency carefully because it brings [a number of other things with it](https://github.com/kristerkari/stylelint-scss/blob/5092e53e447e323d28645eab97f701788e35e87a/package.json#L15-L19) that we may not want, and this one would be very easy to implement. Also, we'd want an autofix, and that rule doesn't seem to have that.

PR #393, opened 2020-02-21
feat(scripts): disallow explicit .scss extension in imports

wincent:
This is just a simple implementation of the rule that doesn't try to handle compound import statements eg.

    @import 'a', 'b', 'c';

because we don't have any of those in liferay-portal. We can add a separate rule to enforce that that invariant remains true, if we want to.

Test plan: Copy this over to liferay-portal and see it make [these changes](https://gist.github.com/wincent/189c5772ae8487b2220a1b638a7bb137).

Closes: https://github.com/liferay/liferay-npm-tools/issues/391


wincent:
Given that this tested fine in liferay-portal, going to merge this and will loop back to address any feedback that comes in later.

PR #395, opened 2020-02-25
feat(scripts): enforce one SCSS import per line

wincent:
Incorrect:

    @import 'a', "b", url(c.css), url('d.css'), url("e.css");
    @import 'f', 'g';

    @import 'first',
            'second',
            url('third.css');

Correct:

    @import 'a';
    @import "b";
    @import url(c.css);
    @import url('d.css');
    @import url("e.css");
    @import 'f';
    @import 'g';

    @import 'first';
    @import 'second';
    @import url('third.css');

As usual, this is relatively straightforward and only the whitespace-handling is tricky. Take the `@import 'first'...` above, for example: because it is preceded by a blank line, if we just clone it and create three imports for `'first'`, `'second'` and `url('third.css')`, then they will each have a blank line before them, and we don't want that.

So, add some special casing in there to detect blank lines and strip them from the 2nd, 3rd (etc) cloned nodes but keep them on the 1st node. This works for the examples in the tests, but I am not trying to handle "crazy" whitespace like trailing tabs because I would expect that Prettier will generally keep that kind of thing out of the codebase.


wincent:
Tests pass and ran this in liferay-portal without it mangling anything, so I think this one is good to go.

Issue #226, opened 2019-09-05
feat: add better support for monorepos to liferay-changelog-generator


wincent:
Right now, we're only using the generator in single-package repos like [alloy-editor](https://github.com/liferay/alloy-editor), or monorepos where we release all the packages together like [liferay-js-themes-toolkit](https://github.com/liferay/liferay-js-themes-toolkit) (where it makes sense to treat changes across all packages indistinctly).

It would be useful to have the generator do useful things in monorepos like this one where things are versioned independently (ie. we often create new releases of liferay-npm-scripts and release them alone). The basic idea would be to have per-package changelogs, and have the generator only include changes that affect a given package directory in the corresponding log.

PR #398, opened 2020-02-29
feat(changelog): add support for per-package monorepo changelogs

wincent:
As an example of what these look like, this commit includes changelogs for the packages in this repo (except for liferay-js-insights, which has apparently never been published), made using:

    (cd packages/liferay-changelog-generator && bin/liferay-changelog-generator.js --version=liferay-changelog-generator/v1.2.0 --to=liferay-changelog-generator/v1.2.0 --no-update-tags --regenerate)
    (cd packages/liferay-jest-junit-reporter && ../liferay-changelog-generator/bin/liferay-changelog-generator.js --version=liferay-jest-junit-reporter/v1.2.0 --to=liferay-jest-junit-reporter/v1.2.0 --no-update-tags --regenerate)
    (cd packages/liferay-js-publish && ../liferay-changelog-generator/bin/liferay-changelog-generator.js --version=liferay-js-publish/v1.0.1 --to=liferay-js-publish/v1.0.1 --no-update-tags --regenerate)
    (cd packages/liferay-npm-bundler-preset-liferay-dev && ../liferay-changelog-generator/bin/liferay-changelog-generator.js --version=liferay-npm-bundler-preset-liferay-dev/v4.2.5 --to=liferay-npm-bundler-preset-liferay-dev/v4.2.5 --no-update-tags --regenerate)
    (cd packages/liferay-npm-scripts && ../liferay-changelog-generator/bin/liferay-changelog-generator.js --version=liferay-npm-scripts/v28.0.1 --to=liferay-npm-scripts/v28.0.1 --no-update-tags --regenerate)

That is a bit of a handful obviously, but subsequent updates should be easy. For example, for the next release of liferay-changelog-generator, we'll just have to run:

    bin/liferay-changelog-generator.js --version=liferay-changelog-generator/v1.3.0

(In other projects, we'd use `npx liferay-changelog-generator` instead of `bin/liferay-changelog-generator.js`.)

Closes: https://github.com/liferay/liferay-npm-tools/issues/226

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,8 +22,16 @@ yarn ci
 # If any checks fail, fix them, submit a PR, and when it is merged,
 # start again. Otherwise...
 
-# Update the version number:
+# Change to the directory of the package you wish to publish:
 cd packages/liferay-npm-scripts
+
+# Update the changelog:
+../liferay-changelog-generator --version=liferay-npm-scripts/v29.0.1
+
+# Review and stage the generated changes:
+git add -p
+
+# Update the version number:
 yarn version --minor # or --major, or --patch
 ```
 
@@ -31,8 +39,10 @@ Running `yarn version` has the following effects:
 
 -   The "preversion" script will run, which effectively runs `yarn ci` again.
 -   The "package.json" gets updated with the new version number.
--   A tagged commit is created.
+-   A tagged commit is created, including the changes to the changelog that you previously staged.
 -   The "postversion" script will run, which automatically does `git push` and performs a `yarn publish`, prompting for confirmation along the way.
+
+Copy the relevant section from the changelog to the corresponding entry on the [releases page](https://github.com/liferay/liferay-npm-tools/releases).
 
 After the release, you can confirm that the packages are correctly listed in the NPM registry:
 
@@ -94,7 +104,10 @@ yarn ci
 # Move into the package's directory
 cd packages/liferay-npm-scripts
 
-# Pick a prerelease version number
+# Update the changelog.
+../liferay-changelog-generator --version=$PACKAGE_NAME/v9.5.0-beta.1
+
+# Bump to the prerelease version number
 yarn version --new-version 9.5.0-beta.1
 
 # Because you are not on the "master" branch,
--- a/packages/liferay-changelog-generator/CHANGELOG.md
+++ b/packages/liferay-changelog-generator/CHANGELOG.md
@@ -2,14 +2,20 @@
 
 [Full changelog](https://github.com/liferay/liferay-npm-tools/compare/liferay-changelog-generator/v1.1.0...liferay-changelog-generator/v1.2.0)
 
+### :new: Features
+
 -   feat: make changelog output identical to Prettier ([\#230](https://github.com/liferay/liferay-npm-tools/pull/230))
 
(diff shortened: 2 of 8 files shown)


wincent:
Note this is only lightly tested, so will want to run it through its paces on a few repos before merging it. I've wanted this for a while now, especially for this repo, because we put out a lot of `liferay-npm-scripts` releases especially, and sometimes people (especially folks from support) want an overview of all the changes, so that they can determine how/whether safe things are to backport.

wincent:
I see some duplicate entries [here](https://github.com/liferay/liferay-npm-tools/blob/f637a63872c63dddfb94822b5eab6b5a3cf48190/packages/liferay-npm-scripts/CHANGELOG.md); will look into that.

wincent:
Going to get this in and will follow up on any feedback later.

bryceosterhaus:
@wincent one thing I noticed the other day with clay and using conventional commits  is that it auto-generated changelogs and new versions when it didnt need to because I didn't pull down the previous tags(or didn't push the tags last time I versioned). `lerna version` looks at the diff based on the last tag, so it made sense that it would auto-generate oddly.

Anyways, not sure that helps give you any more info that you already had. I just noticed a ton of new tags pushed today, so wasn't sure if that was causing you any pains.

wincent:
@bryceosterhaus

>  I just noticed a ton of new tags pushed today, so wasn't sure if that was causing you any pains.

No, that was me. It wasn't new tags, it was just me copying the release notes into previous releases that had never had them ([example](https://github.com/liferay/liferay-npm-tools/releases/tag/liferay-npm-scripts%2Fv25.0.0)).

> it auto-generated changelogs and new versions when it didnt need to because I didn't pull down the previous tags(or didn't push the tags last time I versioned)

That mostly shouldn't happen with `liferay-changelog-generator` because it always [fetches the latest tags](https://github.com/liferay/liferay-npm-tools/blob/29545e04a57b35b17903aa638c7db2be129995ae/packages/liferay-changelog-generator/src/index.js#L616-L623). If somebody else publishes a release but doesn't push tags then you might still not get the latest tags, but `liferay-js-publish` [does a `push` with `--follow-tags`](https://github.com/liferay/liferay-npm-tools/blob/29545e04a57b35b17903aa638c7db2be129995ae/packages/liferay-js-publish/src/index.js#L61), so hopefully that eliminates (or at least mitigates) that source of human error too.

For the most part, all this works pretty well, with the only wrinkle being that some cross-cutting pull requests touch multiple packages and therefore contribute to the changelogs of multiple packages, and the result can be confusing. You have to be thoughtful when deciding how to split things up into pulls. For example, a pull like https://github.com/liferay/liferay-npm-tools/pull/381 is basically kryptonite for a monorepo that uses Conventional Commits:

- It is labeled as a "breaking change" because updating `liferay-npm-scripts` will produce a bunch of new "formatting errors" because trailing commas become required.
- It touches pretty much all the files in the repo because I included https://github.com/liferay/liferay-npm-tools/pull/381/commits/4ac4c198ffcfb18804fa348f6949bd70bfb08172 in that PR ("style: use trailing commas in this repo too").
- So every package in the monorepo gets a confusing "breaking change" in its changelog, unless you have enough context to know what's going on and edit it out.

In conclusion, now that the changelog generator can do per-package changelogs in monorepos, the right way to have prepared #381 would have been to just make the "breaking change" in liferay-npm-scripts, and then do a separate PR of type "style" that applied to the entire repo.

PR #407, opened 2020-03-05
docs: update release instructions

wincent:
I just noticed while preparing the release of liferay-npm-scripts v28.0.2 that I hadn't included the full path to the generator in the instructions; ie.

    ../liferay-changelog-generator/bin/liferay-changelog-generator.js

Let's just simplify the instructions and tell people to use `npx` because that is much easier. We can assume that if you are working on the changelog generator itself and you want to use an unreleased version of the generator then you probably know when not to use `npx` and provide a full path instead.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ yarn ci
 cd packages/liferay-npm-scripts
 
 # Update the changelog:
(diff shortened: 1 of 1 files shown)


wincent:
Just a simple doc update so going to merge this. Created https://github.com/liferay/liferay-npm-tools/issues/408 as a follow-up because that would make releasing even easier.

Issue #411, opened 2020-03-13
Add ".css" to the list of css-loader tests


wincent:
[Here](https://github.com/liferay/liferay-npm-tools/blame/deabf721447dfc2f4ad117ef2a90df09911d1b50/packages/liferay-npm-bundler-preset-liferay-dev/config.json#L272). Because otherwise people end up having to configure it manually in multiple places (eg. [one here now](https://github.com/liferay/liferay-portal/blob/6622a92d19e7414ae9a9e1e02944b21885c5e890/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/.npmbundlerrc#L24-L29), with another inbound soon).

Note that CSS files loaded in this way won't go through the combo loader and that is probably ok if the number or requests remains small (which it definitely will, for now). If this becomes a problem later, we can look at implementing some sort of batching mechanism that leverages the combo loader.

PR #412, opened 2020-03-13
feat(preset): use css-loader for ".css" files

wincent:
In addition to the existing support for ".scss" files.

Closes: https://github.com/liferay/liferay-npm-tools/issues/411

### DIFF
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -269,7 +269,15 @@
 	"output": "build/node/packageRunBuild/resources",
 	"rules": [
 		{
-			"test": "\\.s?css$",
+			"test": "\\.css$",
+			"use": [
+				{
+					"loader": "css-loader"
+				}
+			]
+		},
+		{
+			"test": "\\.scss$",
 			"exclude": "node_modules",
 			"use": [
 				{
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -269,7 +269,7 @@
 	"output": "build/node/packageRunBuild/resources",
 	"rules": [
 		{
-			"test": "\\.scss$",
+			"test": "\\.s?css$",
 			"exclude": "node_modules",
 			"use": [
 				{
--- a/packages/liferay-npm-bundler-preset-liferay-dev/config.json
+++ b/packages/liferay-npm-bundler-preset-liferay-dev/config.json
@@ -270,11 +270,7 @@
 	"rules": [
 		{
 			"test": "\\.css$",
-			"use": [
-				{
-					"loader": "css-loader"
-				}
-			]
+			"use": ["css-loader"]
 		},
 		{
 			"test": "\\.scss$",


wincent:
@izaera: if I understand it correctly, I can use any legal `RegExp` pattern here, and it will get turned into a `RegExp` [here](https://github.com/liferay/liferay-js-toolkit/blob/88d79d4e986a75aea5cc4fddfd76e0afacb42c93/packages/liferay-npm-build-tools-common/src/project/rules.ts#L220), right?

In any case, obviously, I'll actually test this in liferay-portal to make sure it works.

wincent:
Pff, testing this, it does indeed work, but not for the use cases inside liferay-portal because those are inside "node_modules" and we have an `exclude` for that.

My test plan:

Temporarily overwrite preset:

```sh
cp ~/code/liferay-npm-tools/packages/liferay-npm-bundler-preset-liferay-dev/config.json ~modules/node_modules/liferay-npm-bundler-preset-liferay-dev/
```

Delete the project-local config:

```diff
diff --git a/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/.npmbundlerrc b/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/.npmbundlerrc
index aab2c0b45898..28485f36942b 100644
--- a/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/.npmbundlerrc
+++ b/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/.npmbundlerrc
@@ -20,11 +20,5 @@
 	},
 	"ignore": [
 		"**/__tests__/*.js"
-	],
-	"rules": [
-		{
-			"test": "\\.css$",
-			"use": ["css-loader"]
-		}
 	]
 }
\ No newline at end of file
```

Then:

```sh
cd modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type
rm -r build
portool deploy
head -3 build/node/packageRunBuild/resources/node_modules/dynamic-data-mapping-form-field-type\$leaflet@1.2.0/dist/leaflet.css.js
```

If we leave in the `exclude`, the CSS file doesn't get processed (it doesn't get created). If I remove the `exclude`, it does; ie. `head -3` shows:

```js
var link = document.createElement("link");
link.setAttribute("rel", "stylesheet");
link.setAttribute("type", "text/css");
```

Are there performance or other concerns if we drop the `exclude` @izaera? Sadly, keeping it would mean that we still have to keep the local configuration in projects that want to do this, obviating much of the benefit.

In principal, we expect all CSS that is local to liferay-portal to be SCSS, so I guess I could tweak this to be a separate rule (ie. ".scss" excluding "node_modules" and ".css" not excluding it). I'll push that up and we'll see what you think of it.

wincent:
FWIW, I can't see any speed difference. The dynamic-data-mapping-form-field-type build takes about a minute with or without this change.

izaera:
The `exclude` shouldn't be a performance problem (related to the build), because we must test every file in every package for configured loaders so the impact will be minimal.

Activating the "import css" feature for every project in the portal can lead to performance problems in the future during runtime. Think of a page with 20 portlets importing two CSS each: you will end up with 40 CSS requests for a single page.

However, we may fix that in the future if we substitute the `css-loader` by another smart loader that make CSS requests in bunchs. But it is not clear (now) how that would work or the effort needed to do it.

wincent:
Sounds like we're in agreement then. This is what I wrote in #411:

> Note that CSS files loaded in this way won't go through the combo loader and that is probably ok if the number or requests remains small (which it definitely will, for now). If this becomes a problem later, we can look at implementing some sort of batching mechanism that leverages the combo loader.

PR #420, opened 2020-03-24
fix: change contains() to includes()

izaera:
Just a fix


wincent:
Never again!

izaera:
> Never again!

No more fixes allowed? :thinking:

PR #422, opened 2020-03-24
docs(publish): make non-installation requirement clearer

wincent:
By moving it to the top, as suggested [here](https://github.com/liferay/liferay-js-themes-toolkit/pull/451#discussion_r397218823).

And while we're editing, note that we'll soon be using this in liferay-js-themes-toolkit too.


bryceosterhaus:
👍 really like adding more npx examples for cli related tools.

izaera:
LGTM

PR #426, opened 2020-04-01
chore: update bundler dependencies to v2.18.4

izaera:
This is to integrate:

https://github.com/liferay/liferay-js-toolkit/issues/553

into portal's build.


izaera:
Merging this as it is only an upgrade in versions and @wincent gave the OK in a private conversation.

PR #428, opened 2020-04-03
docs(scripts): document how "build" preset values are used

wincent:
I was asking about this in Slack because I initially thought these were all getting passed to `liferay-npm-bundler`, but `liferay-npm-bundler` doesn't even have most of these properties (only `output`).

So, did a bit of digging to see how each value is used and am documenting it for posterity. There is a risk that this might drift out of sync with the implementation if it changes, but I think it's better to have (possibly) slightly stale documentation than nothing at all.


izaera:
LGTM

Issue #416, opened 2020-03-18
Run formatting checks more broadly when liferay-npm-scripts version changes


wincent:
Just an idea: we saw SF fail in [this PR](https://github.com/liferay/liferay-portal-ee/pull/18535) because when we did [this backport](https://github.com/liferay/liferay-portal-ee/pull/18336) of liferay-npm-scripts to the 7.2.x branch, new formatting rules came into effect but we didn't run the checks over the entire repo (in CI, we only run on changed/new files).

The idea, then, is to run more broadly on branches that change the liferay-npm-scripts version (because any updates to eslint-config-liferay — or stylelint, or prettier — end up coming in via a new version of liferay-npm-scripts).

A full-repo formatting check takes about 2m. The top-level package.json doesn't change all that often, so this won't happen often (for example, looking at [the history](https://github.com/liferay/liferay-portal/commits/master/modules/package.json) right now, I see only 8 changes in the last 30 days, and 30 changes over the last 3 months).

If we want to reduce the likelihood of doing a full run unnecessarily, we could also look at the actual diff and detect whether the version of liferay-npm-scripts actually changed.

PR #435, opened 2020-04-07
feat(scripts): don't filter files when liferay-npm-scripts updates

wincent:
As described in #416:

> We saw SF fail in this PR:
>
> https://github.com/liferay/liferay-portal-ee/pull/18535
>
> because when we did this backport:
>
> https://github.com/liferay/liferay-portal-ee/pull/18336
>
> of liferay-npm-scripts to the 7.2.x branch, new formatting rules came into effect but we didn't run the checks over the entire repo (in CI, we only run on changed/new files).
>
> The idea, then, is to run more broadly on branches that change the liferay-npm-scripts version (because any updates to eslint-config-liferay — or stylelint, or prettier — end up coming in via a new version of liferay-npm-scripts).
>
> A full-repo formatting check takes about 2m. The top-level package.json doesn't change all that often, so this won't happen often. For example, looking at the history:
>
> https://github.com/liferay/liferay-portal/commits/master/modules/package.json
>
> right now, I see only 8 changes in the last 30 days, and 30 changes over the last 3 months).
>
> If we want to reduce the likelihood of doing a full run unnecessarily, we could also look at the actual diff and detect whether the version of liferay-npm-scripts actually changed.

In the end that's exactly what I did: we look at the diff to see if it includes added/removed (ie. changed) lines containing the string "liferay-npm-scripts" in either "modules/package.json" or "modules/private/package.json".

Closes: https://github.com/liferay/liferay-npm-tools/issues/416

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
+++ b/packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
@@ -45,6 +45,10 @@ function filterChangedFiles(files) {
 			'modules/private/package.json'
 		);
 	} catch (error) {
+		// An exit status of 1 means we detected the change we were looking for:
+		//
+		// https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---quiet
+		// https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code
 		if (error.toString().includes('exited with status 1.')) {
 			return files;
 		}
--- a/packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
+++ b/packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
@@ -15,6 +15,12 @@ const git = require('./git');
  * to that branch (usually "master", but may also be "master-private").
  *
  * If the variable is not set, the `files` list is return unchanged.
+ *
+ * One important exception to the above: if the top-level `package.json`
+ * changes (which happens rarely), this may indicate a change of the
+ * liferay-npm-scripts version, and in that case we want to run against
+ * the entire unfiltered `files` list.
+ *
  */
 function filterChangedFiles(files) {
 	const upstream = process.env.LIFERAY_NPM_SCRIPTS_WORKING_BRANCH_NAME;
@@ -26,6 +32,23 @@ function filterChangedFiles(files) {
 	const topLevel = git('rev-parse', '--show-toplevel');
 
 	const mergeBase = git('merge-base', 'HEAD', upstream);
+
+	// Check for changes in liferay-npm-scripts version.
+	try {
+		git(
+			'diff',
+			mergeBase,
+			'-Gliferay-npm-scripts',
+			'--quiet',
+			'--',
+			'modules/package.json',
+			'modules/private/package.json'
+		);
+	} catch (error) {
+		if (error.toString().includes('exited with status 1.')) {
+			return files;
+		}
+	}
 
 	const changedFiles = git(
 		'diff',
--- a/packages/liferay-npm-scripts/src/utils/run.js
+++ b/packages/liferay-npm-scripts/src/utils/run.js
@@ -11,7 +11,18 @@ const {spawn} = require('cross-spawn');
 function run(executable, ...args) {
 	const command = [executable, ...args].join(' ');
 
-	const {error, signal, status, stdout} = spawn.sync(executable, args);
+	const {error, signal, status, stderr, stdout} = spawn.sync(
+		executable,
+		args
+	);
+
+	if (error || signal || status) {
+		if (process.env.CI && process.env.TRAVIS) {
+			process.stderr.write('command:\n\n' + command + '\n\n');
+			process.stderr.write('stdout:\n\n' + stdout + '\n\n');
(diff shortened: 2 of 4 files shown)


wincent:
Seeing as path-related stuff is going on here, there's a chance CI might fail on Windows, even though I did my best to write this in a platform-agnostic way. Will find out soon. 😂

packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
```diff
@@ -27,6 +33,23 @@ function filterChangedFiles(files) {
...
+			'diff',
+			mergeBase,
+			'-Gliferay-npm-scripts',
+			'--quiet',
+			'--',
+			'modules/package.json',
+			'modules/private/package.json'
+		);
+	} catch (error) {
+		if (error.toString().includes('exited with status 1.')) {
```
wincent, line 48: Note that we are checking for "status 1." (see the full-stop here). This is to avoid fails positives, like if it exits with a status code of 129 (which can happen if `git-diff` is invoked improperly).

packages/liferay-npm-scripts/src/utils/run.js
```diff
@@ -9,7 +9,7 @@ const {spawn} = require('cross-spawn');
  * Convenience helper for running commands.
  */
 function run(executable, ...args) {
-	const command = `${executable} ${args.join(' ')}`;
+	const command = [executable, ...args].join(' ');
```
wincent, line 12: In adding tests, I saw that we were reporting a failed command like `false` with a trailing space, so this change fixes that.

packages/liferay-npm-scripts/src/utils/run.js
```diff
@@ -19,13 +19,13 @@ function run(executable, ...args) {
...
 		throw new Error(
-			`run(): command \`${command}\` exited due to signal ${signal}`
+			`run(): command \`${command}\` exited due to signal ${signal}.`
 		);
 	}
 
 	if (status) {
 		throw new Error(
-			`run(): command \`${command}\` exited with status ${status}`
+			`run(): command \`${command}\` exited with status ${status}.`
```
wincent, line 28: Here I add the trailing full-stop so that I can do a stricter match, ie. looking for "status 1." as opposed to just "status 1". I could also have used `endsWith()`, but decided to do it this way because it seems good to have a well-defined anchor to match against (for example, if an error ever got dumped in to a log and we wanted to scan inside that, we couldn't use `endsWith()`.

packages/liferay-npm-scripts/test/utils/run.js
```diff
@@ -0,0 +1,20 @@
...
+	it('runs a command and returns its output', () => {
+		expect(run('echo', 'hello')).toBe('hello');
+	});
+
+	it('reports the exit status of failed commands', () => {
+		// NOTE: we want the error message format to be stable, because
+		// `filterChangedFiles()` depends on it.
+		expect(() => run('false')).toThrow(
+			'run(): command `false` exited with status 1.'
+		);
```
wincent, line 18: Added tests here because it is important that the error output remains stable, seeing as we're matching against it. I also thought about assigning the status code as a property on the returned `Error` object, but I didn't want to over-engineer things.

packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
wincent: Further reading:

- https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---quiet
- https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code

I might add this as a code comment, as GitHub comment is probably too ephemeral.

packages/liferay-npm-scripts/src/utils/filterChangedFiles.js
```diff
@@ -27,6 +33,27 @@ function filterChangedFiles(files) {
...
+	// Check for changes in liferay-npm-scripts version.
+	try {
+		git(
+			'diff',
+			mergeBase,
+			'-Gliferay-npm-scripts',
+			'--quiet',
+			'--',
+			'modules/package.json',
+			'modules/private/package.json'
```
wincent, line 45: On repos/branches that don't have `modules/private`, this pathspec is harmless.

wincent:
> Seeing as path-related stuff is going on here, there's a chance CI might fail on Windows, even though I did my best to write this in a platform-agnostic way. Will find out soon. 😂

Aaand of course it failed 😂 — https://travis-ci.org/github/liferay/liferay-npm-tools/jobs/672195382

wincent:
Ok, so it may not be related to paths at all. It could be because `git` isn't installed in the Travis environment. 🤔

https://travis-ci.org/github/liferay/liferay-npm-tools/jobs/672201266

wincent:
Hm, CI seemed wedged. Going to close and re-open to see if that makes a difference.

feat_scripts___don_t_filter_files_when_liferay-npm-scripts_updates_by_wincent_·_Pull_Request__435_·_liferay_liferay-npm-tools

wincent:
Bah, that didn't work either. Will open a new PR.

PR #439, opened 2020-04-17
chore(scripts): update liferay-theme-tasks to 10.0.0

izaera:
No remarks. See commit.


packages/liferay-npm-scripts/package.json
```diff
@@ -45,6 +45,7 @@
 		"eslint-config-liferay": "20.0.1",
 		"fs-extra": "^8.1.0",
 		"globby": "11.0.0",
+		"gulp": "^4.0.2",
```
wincent, line 48: Can you confirm my understanding of this?

This used to be implicit because in portal, liferay-theme-tasks depends on liferay-theme-tasks, which used to depend on gulp-storage, and that pulled in 3.9.1 for us.

Now, because [gulp is only a peer-dep](https://github.com/liferay/liferay-js-themes-toolkit/blob/liferay-theme-tasks/v10.0.0/packages/liferay-theme-tasks/package.json), we need to do this explicitly in order for portal to have a copy of Gulp on-hand.

Right?

wincent: Ignore this one because I actually read your commit message now... 😂

izaera: Yes. The output from `yarn why gulp` in `liferay-portal` with theme-tasks v9 shows:

```shell
$ yarn why gulp
yarn why v1.22.4
[1/4] Why do we have the module "gulp"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "gulp@3.9.1"
   - "_project_#liferay-npm-scripts#liferay-theme-tasks#gulp-storage" depends on it
   - Hoisted from "_project_#liferay-npm-scripts#liferay-theme-tasks#gulp-storage#gulp"
info Disk size without dependencies: "408KB"
info Disk size with unique dependencies: "408KB"
info Disk size with transitive dependencies: "408KB"
info Number of shared dependencies: 130
Done in 2.03s.
```

izaera:
done

Issue #444, opened 2020-04-30
Override Prettier line-comment placement at start of blocks


wincent:
Originally created [here](https://github.com/liferay/eslint-config-liferay/issues/178) but then realized that it best belongs in this repo because I'm not going to make any changes over there. Copy-paste follows:

---

As [I explained](https://github.com/brianchandotcom/liferay-portal/pull/88193) when I shipped [the lint to enforce blank lines before and after comments](https://github.com/liferay/eslint-config-liferay/issues/169):

> This gets up 95% of the way to where we want to go, by which I meant that:
> 
> * We put blank lines before and after line comments (`//`); and:
> * Blank lines before but not after block comments (`/* ... */`)
> 
> in 95% of places. The remaining 5% consists of points where ESLint and Prettier disagree about how a comment should be formatted; this specifically occurs at the beginning of objects, arrays, blocks etc. The first 95% was cheap to get (configuration only); the remaining 5% would be expensive (requiring hacking Prettier) so I stopped at this point for now.

I don't think Brian read that explanation, because he effectively [asked us to deal with the remaining 5%](https://github.com/brianchandotcom/liferay-portal/commit/1f085b085dd2b1260bb4da32229c060a08bb3363) as well anyway, so I'm creating this ticket for that.

For prioritization purposes, the usual observations about overriding Prettier apply:

- It's not necessarily cheap to do so, so we shouldn't rush to do it.
- If it involves changes that would conflict with Prettier's wrapping algorithm, then we probably shouldn't do it at all. [This is an example of the explanation](https://github.com/brianchandotcom/liferay-portal/pull/87924#issuecomment-618875826) for that:

> Custom rules that override Prettier's wrapping choices. In general, deciding where to wrap is the "one job" that Prettier has. It uses a [neat algorithm](https://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) that is "optimal" (minimizes line breaks) and "bounded" (uses a finite amount of lookhead, in this case, a line-width of lookahead). Where a line breaks can have a cascading effect on where other lines break. We cannot change this logic without forking, except for trivial cases like forcing `else` and `catch` onto a different line like we did for JS (_that_ tweak produces good results that don't significantly compromise the algorithm only because they change the line with by just two chars).

[...]

PR #445, opened 2020-04-30
feat(scripts)!: make "blank line before comments" follow Java convention

wincent:
We were doing this in most places thanks to [this change in eslint-config-liferay](https://github.com/liferay/eslint-config-liferay/issues/169), but there were a few places where we didn't enforce blank lines because ESLint and Prettier disagree about comments at the beginning and end of blocks, classes, arrays, objects etc.

Brian would like us to handle those "few places" too, so after thinking about it for a bit, I realized that it is actually quite easy because:

1. We already have this post-Prettier hook in place in liferay-npm-scripts that allows us to override Prettier if needed.
2. We don't need to make a custom lint rule for it; we can just use the existing "lines-around-comment" rule that ESLint provides, appropriately configured.
3. This doesn't fundamentally undermine the main mission of Prettier, which is to wrap lines, because inserting/removing blank lines never changes line width.

I'm feeling pretty lucky that I used all those crazy Unicode sentinels when I implemented JSP formatting, because it means that we can ignore all the temporary placeholder comments that get inserted into the source code so that Prettier can parse it. eg.

```jsp
<c:if test="<% blah() %>">
   stuff();
</c:if>
```

gets transformed into (something like):

```js
// crazy Unicode placeholder
  stuff();
// crazy Unicode placeholder
```

to make it into valid JS. Now, with this change, Prettier + ESLint would want to inject blank lines:

```js
// crazy Unicode placeholder

  stuff();

// crazy Unicode placeholder
```

and when we round-trip back to JSP again, it would be:

```jsp
<c:if test="<% blah() %>">

   stuff();

</c:if>
```

which is obviously not what we want. Luckily for us, we can configure ESLint to ignore the "crazy Unicode stuff", and that means we don't mangle any of our JSP.

Test plan: Run on liferay-portal, producing [this diff](https://gist.github.com/wincent/6ff619293a0644a731c621b6aa2819ed).

Closes: https://github.com/liferay/liferay-npm-tools/issues/444

### DIFF
--- a/packages/liferay-npm-scripts/src/jsp/getPaddedReplacement.js
+++ b/packages/liferay-npm-scripts/src/jsp/getPaddedReplacement.js
@@ -47,5 +47,7 @@ function getPaddedReplacement(match, template) {
 }
 
 getPaddedReplacement.IDENTIFIER = IDENTIFIER;
+getPaddedReplacement.ID_END = ID_END;
+getPaddedReplacement.ID_START = ID_START;
 
 module.exports = getPaddedReplacement;
--- a/packages/liferay-npm-scripts/src/jsp/tagReplacements.js
+++ b/packages/liferay-npm-scripts/src/jsp/tagReplacements.js
@@ -141,6 +141,8 @@ function getSelfClosingTagReplacement(tag) {
 }
 
 module.exports = {
+	BLOCK_CLOSE,
+	BLOCK_OPEN,
 	CLOSE_TAG,
 	OPEN_TAG,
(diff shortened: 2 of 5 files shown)


wincent:
This is a relatively simple change and as you can see in the test plan, works fine in liferay-portal, so I'm just going to merge it, now that CI is green.

wincent:
Testing PR for liferay-portal [here](https://github.com/wincent/liferay-portal/pull/244).

PR #447, opened 2020-05-07
chore: update to eslint-config-liferay v21.1.0

wincent:
Not a breaking change because the added rule is not on by default.

Release notes:

https://github.com/liferay/eslint-config-liferay/releases/tag/v21.1.0


wincent:
Just a routine update so going to merge it.

PR #458, opened 2020-06-02
docs(scripts): fix version numbers in CHANGELOG.md

wincent:
Caused because I thought the next version was going to be v33.3.0 but then when I actually updated/published (with `yarn version --minor`) I saw that it's actually v32.3.0.


wincent:
Trivs.

PR #459, opened 2020-06-03
chore(preset): add @clayui/layout package to preset

wincent:
[This was released two weeks ago](https://www.npmjs.com/package/@clayui/layout) and we're ready to start using it in liferay-portal:


wincent:
If it's green, it's clean.

wincent:
Yeah, it's easy to overlook, because we don't add new packages that often.

The purpose of this is just so that people can `yarn add @clayui/blah` to their `package.json` in `modules/apps/some-app/some-app-web/` project and it will use the shared version that you added to `frontend-taglib-clay` instead of them getting their own unique copy of it bundled into their project.

Issue #467, opened 2020-06-18
Catch whole-script indentation problems in JSP


wincent:
As seen in this PR: https://github.com/brianchandotcom/liferay-portal/pull/90231

The way our JS-in-JSP formatting works at the moment is the following:

- Extract JS.
- Pass through Prettier.
- Put JS *back where it was*.

This means that if the *entire* block is misindented (ie. too far to the left or the right) but *each line* is correctly indented relative to its neighbors, the formatting won't complain, because the round trip via Prettier won't actually change anything.

So, we need to add an additional layer of checking here, which is that the initial indent is correct compared to its containing block.

PR #468, opened 2020-06-18
fix(scripts): catch whole-script indentation problems in JSP (#467)

wincent:
The way our JS-in-JSP formatting works is as follows:

1. Extract JS.
2. Pass through Prettier.
3. Put JS back where it was.

This means that if the entire block is misindented (ie. too far to the left or the right) but each line is correctly indented relative to its neighbors, the formatting won't complain, because the round trip via Prettier won't actually change anything.

So, we need add an additional layer of enforcement here, which is that the initial indent is correct compared to its containing block.

The fix is straightforward, but I also wanted to make sure that it played nicely in an edge case scenario like this:

    <script>alert('boom');</script>

We don't have any tags like that in liferay-portal, but if we did we wouldn't want to mess them up. In general, the Java source formatter wants opening and closing tags to be on lines of their own. So I made it reformat them to:

    <script>
        alert('boom');
    </script>

Closes: https://github.com/liferay/liferay-npm-tools/issues/467

### DIFF
--- a/packages/liferay-npm-scripts/src/jsp/dedent.js
+++ b/packages/liferay-npm-scripts/src/jsp/dedent.js
@@ -85,9 +85,7 @@ function dedent(input, tabWidth = 4) {
 		dedented.pop();
 	}
 
-	const lastMinimum = isFinite(minimum) ? Math.floor(minimum / tabWidth) : 0;
-
-	return [dedented.join('\n'), lastMinimum];
+	return dedented.join('\n');
 }
 
 module.exports = dedent;
--- a/packages/liferay-npm-scripts/test/jsp/__snapshots__/formatJSP.js.snap
+++ b/packages/liferay-npm-scripts/test/jsp/__snapshots__/formatJSP.js.snap
@@ -1229,7 +1229,7 @@ exports[`formatJSP() formatting entire fixtures page.jsp matches snapshot 1`] = 
 
 	<aui:script>
 		window.foo = 'foo';
-</aui:script>
+	</aui:script>
 
 	<aui:script>
 		var SOME_OBJ = {
@@ -1245,7 +1245,7 @@ exports[`formatJSP() formatting entire fixtures page.jsp matches snapshot 1`] = 
 	</aui:script>
 
 	<aui:script require=\\"\\">
-</aui:script>
+	</aui:script>
 </body>
 </html>
 "
--- a/packages/liferay-npm-scripts/test/jsp/formatJSP.js
+++ b/packages/liferay-npm-scripts/test/jsp/formatJSP.js
@@ -88,6 +88,115 @@ describe('formatJSP()', () => {
 		expect(formatJSP(source)).toBe(expected);
 	});
 
+	describe('fixing problems with indentation relative to script tag (#437)', () => {
+		// ie. each line is correct with respect to its neighbors, but overall,
+		// the code is wrong relative to the script tag.
+
+		it('corrects script content that is insufficiently indented', () => {
+			const source = `
+				<aui:script>
+				function <portlet:namespace />deleteOrganization(
+					organizationId,
+					organizationsRedirect
+				) {
+					<portlet:namespace />doDeleteOrganization(
+						'<%= Organization.class.getName() %>',
+						organizationId,
+						organizationsRedirect
+					);
+				}
+				</aui:script>
+			`;
+
+			const expected = `
+				<aui:script>
+					function <portlet:namespace />deleteOrganization(
+						organizationId,
+						organizationsRedirect
+					) {
+						<portlet:namespace />doDeleteOrganization(
+							'<%= Organization.class.getName() %>',
+							organizationId,
+							organizationsRedirect
+						);
+					}
+				</aui:script>
+			`;
+
+			expect(formatJSP(source)).toBe(expected);
+		});
+
+		it('corrects script content that is excessively indented', () => {
+			// Note that it works for <script> as well.
+
+			const source = `
+				<script>
+						function <portlet:namespace />deleteOrganization(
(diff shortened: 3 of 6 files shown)


packages/liferay-npm-scripts/test/jsp/__snapshots__/formatJSP.js.snap
```diff
@@ -1229,7 +1229,7 @@ exports[`formatJSP() formatting entire fixtures page.jsp matches snapshot 1`] =
 
 	<aui:script>
 		window.foo = 'foo';
-</aui:script>
+	</aui:script>
```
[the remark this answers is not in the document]
wincent: So it seems, but because it was tucked away in a snapshot didn't notice.

packages/liferay-npm-scripts/test/jsp/formatJSP.js
```diff
@@ -88,6 +88,115 @@ describe('formatJSP()', () => {
...
 	});
 
+	describe('fixing problems with indentation relative to script tag (#437)', () => {
+		// ie. each line is correct with respect to its neighbors, but overall,
+		// the code is wrong relative to the script tag.
+
+		it('corrects script content that is insufficiently indented', () => {
+			const source = `
+				<aui:script>
+				function <portlet:namespace />deleteOrganization(
```
jbalsas, line 98: I haven't checked the code, I'm sure it's likely covered and that it's highly unlikely we have code like this anywhere... but have you checked with negatively indented code?

```jsp
        <aui:script>
function foo() {
}
        </aui:script>
```

packages/liferay-npm-scripts/src/jsp/dedent.js
```diff
@@ -85,9 +85,7 @@ function dedent(input, tabWidth = 4) {
 		dedented.pop();
 	}
 
-	const lastMinimum = isFinite(minimum) ? Math.floor(minimum / tabWidth) : 0;
-
-	return [dedented.join('\n'), lastMinimum];
```
wincent, line 90: So this was the problem. Previously we just detected the indent of the script contents and trusted that it was ok.

The new approach is to just skip the detection and let the opening `<script>` or `<aui:script>` tag call the shots.

packages/liferay-npm-scripts/test/jsp/formatJSP.js
jbalsas: And by I haven't checked I mean I can barely understand it

wincent: I haven't checked it but I'm sure it works. I'll add a test.

packages/liferay-npm-scripts/test/jsp/__snapshots__/formatJSP.js.snap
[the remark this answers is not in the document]
jbalsas: Double fix then!! 👏

packages/liferay-npm-scripts/test/jsp/formatJSP.js
wincent: Negative indent test: https://github.com/liferay/liferay-npm-tools/pull/468/commits/0ed44061c85fb697878f4282c00802dc2bc5c878

packages/liferay-npm-scripts/test/jsp/formatJSP.js
```diff
@@ -162,6 +162,40 @@ describe('formatJSP()', () => {
 			expect(formatJSP(source)).toBe(expected);
 		});
 
+		it('corrects script content that is "negatively" indented', () => {
```
jbalsas, line 165: I stand answered :)

packages/liferay-npm-scripts/test/jsp/__snapshots__/formatJSP.js.snap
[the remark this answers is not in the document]
wincent: Although, note that this is in Nate's crazy fixture file that came across from the old `csf`. Goodness knows what's going on in this template.

packages/liferay-npm-scripts/test/jsp/formatJSP.js
wincent: > I can barely understand it just by looking over it briefly

Added some explanation in https://github.com/liferay/liferay-npm-tools/pull/468/commits/77351cebc8a099b400fb1704ba602e18978c06cc which might help?

I think the `baseIndent` calculation is the important bit.

The tricky dealing with `prefix` and `suffix` is another story that I can only explain via Jedi mind trick. Nothing to see there... move along...

not-the-droids

wincent:
Ninja-ing, because ninja's are cool.

cool-ninja

Travis seems dead, but `yarn ci` works.

packages/liferay-npm-scripts/test/jsp/formatJSP.js
jbalsas: > [...] plus 1 because columns are
1-indexed [...]

This is pure evil 😂

wincent: Are you saying there's a Sith connection here? It's pretty standard for editors/IDE and tools that operate on source (eg. ESLint, Babel etc) to be 1-based.

evil-sith

jbalsas: > It's pretty standard for editors/IDE [...]

You would know, you VIM lover!

wincent: I once had to open VSCode to help a Flash developer, so I have some experience with that too... 😂

PR #472, opened 2020-07-08
fix(changelog): consistently autodetect prefixes for non-monorepos too

wincent:
Noticed while looking at [the liferay-ckeditor releases](https://github.com/liferay/liferay-ckeditor/releases) that the links to the "full changelog" are missing the "v" prefix in a lot of places.

That's because the code we have to read the prefix from the ".yarnrc" was inadequately tested outside of monorepos. This small fix makes sure that it reads the top-level ".yarnrc" as well, for the non-monorepo case.

Test plan: Make a merge commit in liferay-ckeditor and run the generator with and without a "v" prefix. Then apply this change and repeat the test; results:

1. "master", with "v" prefix: get valid links in changelog.
2. "master", without "v" prefix: get bad links in changelog.
3. This commit, with "v" prefix: get valid links in changelog.
4. This commit, without "v" prefix: get valid links in changelog.

ie. broken case "2" is fixed.

### DIFF
--- a/packages/liferay-changelog-generator/src/index.js
+++ b/packages/liferay-changelog-generator/src/index.js
@@ -632,16 +632,21 @@ async function generate({date, from, remote, to, version}) {
  *     monorepo and a ".yarnrc" file exists, we return its "version-tag-prefix".
  *
  *     For example, given a prefix of "my-package/v", then we can find matching
- *     tags using `git-describe`--match='my-package/v*'".
+ *     tags using `git describe --match='my-package/v*'.
  *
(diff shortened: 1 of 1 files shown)


wincent:
Got incorrectly autoclosed because of link in [commit in other repo](https://github.com/liferay/liferay-ckeditor/commit/e5afc96b66fa1aafe95a7b1594b9b2222ab65ea1)... silly GitHub.

wincent:
Seems like a good candidate for a Travolta merge.

travolta-merge

jbalsas:
👀

PR #473, opened 2020-07-08
docs(changelog): fix broken link in README

wincent:
Because in that project, we switched from top-level changelog to per-package ones.


wincent:
Trivial doc fix, so merging.

PR #481, opened 2020-08-25
fix(scripts): don't merge Babel options into webpack loaders other than babel-loader

brunobasto:
And because these loaders now seem to validate their configuration schema, it will throw errors and fail to build:

```
ERROR in ./src/main/resources/META-INF/resources/styles/main.scss
Module build failed (from /Users/brunobasto/Projects/liferay-portal/modules/node_modules/style-loader/dist/cjs.js):
ValidationError: Invalid options object. Style Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'overrides'. These properties are valid:
   object { injectType?, attributes?, insert?, base?, esModule? }
    at validate (/Users/brunobasto/Projects/liferay-portal/modules/node_modules/style-loader/node_modules/schema-utils/dist/validate.js:85:11)
    at Object.loader (/Users/brunobasto/Projects/liferay-portal/modules/node_modules/style-loader/dist/index.js:22:28)
 @ ./test/dev/components/Summary.js 3:0-74
```

In this fix I'm not merging the babel loader configs if it's not the babel loader. I've also went from changing the values by reference to returning a new merged reference. That fixed an issue where [this line](https://github.com/liferay/liferay-npm-tools/blob/master/packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js#L38) was effectively doing nothing when the `use` was not an array.

### DIFF
--- a/packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
+++ b/packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
@@ -34,10 +34,7 @@ function mergeBabelLoaderOptions(webpackConfig) {
 				}
 
 				const mergeUseEntry = (useEntry) => {
-					if (
-						typeof useEntry === 'string' &&
-						useEntry === 'babel-loader'
-					) {
+					if (useEntry === 'babel-loader') {
 						return {
 							loader: useEntry,
 							options: {...BABEL_CONFIG},
--- a/packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
+++ b/packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
@@ -22,30 +22,45 @@ function mergeBabelLoaderOptions(webpackConfig) {
 		return webpackConfig;
 	}
 
-	webpackConfig.module.rules.forEach((rule) => {
-		let {use} = rule;
+	return {
+		...webpackConfig,
+		module: {
+			...webpackConfig.module,
+			rules: webpackConfig.module.rules.map((rule) => {
+				const {use} = rule;
 
-		if (!use) {
-			return;
-		}
+				if (!use) {
+					return rule;
+				}
 
-		if (!Array.isArray(use)) {
-			use = [use];
-		}
+				const mergeUseEntry = (useEntry) => {
+					if (
+						typeof useEntry === 'string' &&
+						useEntry === 'babel-loader'
+					) {
+						return {
+							loader: useEntry,
+							options: {...BABEL_CONFIG},
+						};
+					} else if (useEntry.loader === 'babel-loader') {
+						return {
+							...useEntry,
+							options: {...BABEL_CONFIG, ...useEntry.options},
+						};
+					}
 
-		use.forEach((useEntry, i) => {
-			if (typeof useEntry === 'string') {
-				use[i] = {
-					loader: useEntry,
-					options: {...BABEL_CONFIG},
+					return useEntry;
 				};
-			} else {
-				use[i].options = {...BABEL_CONFIG, ...useEntry.options};
-			}
-		});
-	});
 
-	return webpackConfig;
+				return {
+					...rule,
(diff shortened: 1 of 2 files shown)


jbalsas:
Love that this has new tests!!

I'll leave this to @wincent when he gets back to vet and merge... this shouldn't block commerce in their migration anyways...

jbalsas:
@bryceosterhaus, feel free to take a look and if you and @brunobasto are satisfied with this, take it for a spin and see how it looks over at DXP ;)

brunobasto:
> @bryceosterhaus, feel free to take a look and if you and @brunobasto are satisfied with this, take it for a spin and see how it looks over at DXP ;)

Impact should be minimal since this only touches the webpack script and in DXP its usage is not widespread.

jbalsas:
> Impact should be minimal since this only touches the webpack script and in DXP its usage is not widespread.

DUH!! You're right!!

packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
```diff
@@ -22,30 +22,45 @@ function mergeBabelLoaderOptions(webpackConfig) {
...
-			use = [use];
-		}
-
-		use.forEach((useEntry, i) => {
-			if (typeof useEntry === 'string') {
-				use[i] = {
-					loader: useEntry,
-					options: {...BABEL_CONFIG},
+	return {
+		...webpackConfig,
```
bryceosterhaus, line 26: We might also want to add the ability for multiple webpack configs, meaning it could be an array of webpack configs.

packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
```diff
@@ -22,30 +22,45 @@ function mergeBabelLoaderOptions(webpackConfig) {
...
+				const {use} = rule;
+
+				if (!use) {
+					return rule;
+				}
+
+				const mergeUseEntry = (useEntry) => {
+					if (
+						typeof useEntry === 'string' &&
+						useEntry === 'babel-loader'
```
wincent, line 39: You don't need the `typeof` check because your `=== 'babel-loader'` already effectively includes such a check.

packages/liferay-npm-scripts/src/utils/mergeBabelLoaderOptions.js
```diff
@@ -22,30 +22,45 @@ function mergeBabelLoaderOptions(webpackConfig) {
...
+					) {
+						return {
+							loader: useEntry,
+							options: {...BABEL_CONFIG},
+						};
+					} else if (useEntry.loader === 'babel-loader') {
```
wincent, line 45: And here `useEntry` could still be a string (harmless enough, I guess, because `someString.loader` isn't going to blow up).

packages/liferay-npm-scripts/test/utils/mergeBabelLoaderOptions.js
```diff
@@ -0,0 +1,83 @@
...
+						test: /\.(js|jsx)$/,
+					},
+				],
+			},
+		};
+
+		expect(mergeBabelLoaderOptions(config)).toEqual(config);
+	});
+
+	it('returns unchanged rules the loader is not the babel-loader', () => {
```
wincent, line 39: "rules the loader" → "rules when the loader"

wincent:
Just going to push a couple small things to address those comments and then we can merge.

wincent:
[CI is green](https://travis-ci.org/github/liferay/liferay-npm-tools/jobs/721739199), so going to pull the trigger on this one.

Like @bryceosterhaus says...

> just one idea for adding support for multiple webpack configs by checking if its an array instead of an object. This could be done in a separate ticket though

...if and when we see that we need to handle arrays of configs, we can add that in a future PR, but for now we can defer taking on any additional complexity.

brunobasto:
Awesome! thx @wincent

Issue #496, opened 2020-09-14
Include @testing-library/react-hooks library


victorg1991:
Could we include the `@testing-library/react-hooks` library to be used within the tests? https://github.com/testing-library/react-hooks-testing-library

We had some misunderstandings and the library is already included in the portal workspace because it is added in [portal-workflow-metrics-web package.json](https://github.com/liferay/liferay-portal/blob/f86d976d9f59933928199ee8b57ebee1c88efbd7/modules/dxp/apps/portal-workflow/portal-workflow-metrics-web/package.json#L36) (using a SF supression) and used [here](https://github.com/liferay/liferay-portal/blob/fe37327dbe1e88de73f83075d7d18ee91c00dc26/modules/apps/layout/layout-content-page-editor-web/test/page_editor/app/components/undo/useUndo.test.js)

To avoid disabling the current test I think it would be better for [this pr](https://github.com/liferay-frontend/liferay-portal/pull/384) to wait until we include the library 'officially' so that we can remove the supression while the tests continue working

wincent:
I'm testing a release of the script with this new dependency today.

wincent:
I cut a release and folded it into [this PR](https://github.com/liferay-frontend/liferay-portal/pull/384), so follow along there if you want to track when this gets into liferay-portal.

PR #499, opened 2020-09-18
feat(scripts): add testing-library/react-hooks

wincent:
We were using v1.0.4 unofficially in liferay-portal due to [this `devDependency`](https://github.com/liferay/liferay-portal/blob/f86d976d9f59933928199ee8b57ebee1c88efbd7/modules/dxp/apps/portal-workflow/portal-workflow-metrics-web/package.json#L36).

We want to make it official and integrate it into liferay-portal "The Right Way™", which means bring it into liferay-npm-scripts. The question is, which version? The latest release is in the v3 range, which means there have been two breaking changes.

In this commit, we try out the latest and we'll see what breaks. During the install `yarn` prints:

    warning " > @testing-library/react-hooks@3.4.1" has unmet peer dependency "react@>=16.9.0".
    warning " > @testing-library/react-hooks@3.4.1" has unmet peer dependency "react-test-renderer@>=16.9.0".

First one is ok because we'll get that from the copy hoisted from frontend-js-react-web in liferay-portal anyway (we have a very loose requirement specified in liferay-npm-scripts' package.json so that we don't wind up pulling in another version by mistake).

Second will be ok too because I am adding it as an explicit dep; again with a loose version requirement because we want yarn to have maximum flexibility to resolve the dependency to a single, unique version.

Test plan:

1. `yarn add` this in liferay-portal.
2. Grab https://github.com/liferay-frontend/liferay-portal/pull/384 and edit it; see that tests still work.

Closes: https://github.com/liferay/liferay-npm-tools/issues/496


wincent:
Still carrying out the test plan. I'll report back here with the results.

packages/liferay-npm-scripts/package.json
```diff
@@ -65,6 +66,7 @@
 		"prettier": "2.0.5",
 		"react": "*",
 		"react-dom": "*",
+		"react-test-renderer": "*",
```
wincent, line 69: Seeing this is a Facebook-provided library, I've gone with the loose version range like we do for `react` and `react-dom` themselves.

wincent:
Testing in liferay-portal:

### `portal-workflow-metrics-web`

```
Test Suites: 72 passed, 72 total
Tests:       259 passed, 259 total
Snapshots:   0 total
Time:        25.586s
Ran all test suites.
✨  Done in 28.52s.
```

These tests have huge amounts of console spew, although they all pass. I think the team should try to address that, and if they need help Frontend Infra can provide it; but it's not a blocker for any of these PRs.

### `layout-content-page-editor-web`

```
Test Suites: 46 passed, 46 total
Tests:       1 todo, 227 passed, 228 total
Snapshots:   0 total
Time:        25.392s
Ran all test suites.
✨  Done in 28.35s.
```

wincent:
Shipping this on the ground that it is `#NoLogicChanges` and I tested it. 😀

PR #500, opened 2020-09-18
feat(scripts): catch more dev dependencies hiding in non-dev section

wincent:
As per [this comment](https://github.com/liferay-frontend/liferay-portal/pull/384/files#r490801624), in that PR, somebody is avoiding the lint warning about adding `devDependencies`, which we don't allow in liferay-portal for reasons [described here](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/dev_dependencies.md), by including them in `dependencies` instead. We can catch that by adding a couple more entries to the blacklist. Obviously this is whack-a-mole, but it's way better than having dev dependencies shoved into built JARs.

Test plan: `yarn add` in liferay-portal and see if our `check` script catches the issue on the PR linked above.

### DIFF
--- a/packages/liferay-npm-scripts/src/presets/standard/index.js
+++ b/packages/liferay-npm-scripts/src/presets/standard/index.js
@@ -42,7 +42,11 @@ module.exports = {
 	check: CHECK_AND_FIX_GLOBS,
 	fix: CHECK_AND_FIX_GLOBS,
 	rules: {
-		'blacklisted-dependency-patterns': ['^liferay-npm-bundler-loader-.+'],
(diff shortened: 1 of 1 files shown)


wincent:
Tested in liferay-portal, and we see the expected error on `yarn checkFormat`:

```
Preflight check failed:
package.json: BAD - contains blacklisted dependency: @testing-library/react-hooks
package.json: BAD - contains blacklisted dependency: react-test-renderer
```

wincent:
Shipping this on the ground that it is `#NoLogicChanges` and I tested it. 😀

